PR #24, opened 2019-08-09
add SanLib entity model plugin

SanAndreasP:


JannisX11:
Is SanLib a publicly used model loader? Could the plugin be valuable for Blockbench users? If this is just a private project, I recommend to load the plugin from your local drive or share it via direct message.

SanAndreasP:
I mean, it is [publicly available](https://github.com/SanAndreasP/SanLib/blob/1.12/src/main/java/de/sanandrew/mods/sanlib/lib/client/ModelJsonLoader.java) for modders to use, if that counts. 😄 

If not, feel free to close this. I'll just put the plugin into the SanLib repo then.

PR #63, opened 2020-05-11
Added Startup Tips Plugin

TheOtterlord:
The Startup Tips plugin provides helpful tips about Blockbench on startup. 

This is a simple plugin, but one that would definitely be helpful to beginners. 
It has a good number of tips already, but feel free to add more.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -183,5 +183,14 @@
 		"version": "1.0.2",
 		"variant": "both",
 		"min_version": "3.3.0"
+	},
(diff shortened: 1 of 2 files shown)


JannisX11:
I would recommend to put the "Show Tip" action into the Help menu, and not in the Filter menu as it's not really a Filter.

TheOtterlord:
@JannisX11 Yeah, good idea. I'll release an update soon

PR #76, opened 2020-07-07
Add multi-layer export plugin

aidancbrady:
Right now it just supports solid & translucent, I don't see a real reason why anyone would need to apply the other layers.


JannisX11:
Hi, please add your plugin metadata to the `plugins.json` file as well.

aidancbrady:
Should be all set.

PR #82, opened 2020-07-24
Fix minor bug in animation_utils plugin exporter

fadookie:
- Start refactoring plugin to use webpack and source modules
- Bump version to 2.0.1


JannisX11:
I don't see a problem with implementing the /src/ folder for plugins. Just wondering about the gitignore file. I think there should be a simple, less bloated global gitignore and then everyone can have their own gitignore setup in their src folder.

Also, wow that's a lot of dependencies in the bundled file. Is that lodash? The file size is like almost half of the main Blockbench source code.

fadookie:
Huh, I am cherry picking a single function from lodash. I was hoping webpack would do tree shaking automatically but maybe I need to turn that on somehow.

fadookie:
@JannisX11 switching from development to production webpack mode reduced bundle from 584K to 92K. cherry-picking lodash function further reduced bundle size from 27K. Is this acceptable?

PR #86, opened 2020-08-02
Animation Utils: Fix 1.0/Bedrock Import

fadookie:
A user reported a bad bug where geckolib 1.0 and bedrock animation keyframe data was not loading properly. This has been fixed.

I also forgot to bump my version in the manifest last time, this is fixed.


src/animation_utils/index.js
```diff
@@ -276,7 +276,8 @@ import software.bernie.geckolib.animation.render.AnimatedModelRenderer;`;
       return result;
   }
 
-  function keyframeExtend(data) {
+  function keyframeExtend(dataIn) {
+    const data = Object.assign({}, dataIn);
```
fadookie, line 280: Not strictly related to bugfix but I realized I was mutating the object passed in which seemed like bad form.

fadookie:
Thanks for the merge!

PR #87, opened 2020-08-03
Model Information

RyanGar46:
Finds out information about your model. more features will be added soon.


JannisX11:
This functionality already exists in Blockbench. You can view statistics by clicking on the cube counter above the outliner. It is also more accurate since disabled faces are not counted.

PR #102, opened 2020-10-14
set max_version for GeckoLib animation utils to 3.6.99

fadookie:
This will prevent users from running into broken, buggy behavior on bb 3.7 until we have time to push an update that fixes those bugs.


JannisX11:
Max version should probably be set to 3.6.99 to prevent users from installing it in 3.7.

fadookie:
@JannisX11 Good to know, I've made the change.

It might be easier though if we could just specify a semver range for compatibility as is the standard in the greater JS ecosystem. Ex. `"required_version": "^3.6"` or `3.6x || 3.7.x` etc. The [semver](https://www.npmjs.com/package/semver) package can handle parsing and comparing versions and ranges.

PR #109, opened 2020-11-12
Adding Triangle Generator plugin

DavidEMorales:
This plugin generates triangles, namely isoceles and right triangles as well as pyramids.


JannisX11:
Sorry for the late reply. Please check these points:

* New plugins should follow the new plugin format using Plugin.register and actions, as described on https://documentation.blockbench.net
* As an example for a plugin that is written using the up-to-date API, check out Plaster
* You also need to add the plugin to the `plugins.json` file, also described in the documentation

PR #111, opened 2020-12-02
Add Texture Stitcher plugin

mchorse:
Hi, so I decided to write a simple texture stitcher plugin. The purpose of it is to be used when converting models to different formats that support single texture. For example, a person made a model using multiple textures, and wants to convert it to GeckoLib format, however, GeckoLib's code supports only one texture. 

The implementation is sufficient ~~pretty naive, I could look into a better algorithm later, but~~ (due to constraints' of project's texture width/height, the implementation is already optimal, and can't be optimized, if I understand correctly how UV mapping works in Blockbench). 😅 

Let me know if you'll have any issues with the code formatting, or the plugin. Also, thanks for the help!

P.S.: Is it me or did CI messed up something? 😕

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -267,5 +267,13 @@
 		"description": "This plugin renames duplicate bones so they work in bedrock and GeckoLib models",
 		"version": "1.0.0",
 		"variant": "both"
+	},
+	"texture_stitcher": {
+		"title": "Texture Stitcher",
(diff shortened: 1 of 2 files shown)


JannisX11:
I think the CI had a connection issue or something

mchorse:
Ayyy, thanks Jannis! 😊

PR #116, opened 2020-12-15
Update csmodel.js

dragonmaster95:
-Imports the UVs in the correct size now instead of basing the UV off of the size of the element.

### DIFF
--- a/plugins/csmodel.js
+++ b/plugins/csmodel.js
@@ -517,9 +517,8 @@ Plugin.register('csmodel', {
 
 				var tex = new Texture({name: 'texture'}).fromDataURL(url).add();
 				tex.load_callback = function() {
-					ratio = tex.height/tex.width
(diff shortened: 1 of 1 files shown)


JannisX11:
Why does it set the texture size to 16 instead of the size of the texture? That seems counter intuitive for editing

dragonmaster95:
It doesn't set it the texture size to 16, it sets the Project resolution to 16x16.
If it bothers anyone they can set the project resolution to a different value.

JannisX11:
I'd like to set this to the right value by default.

dragonmaster95:
Well "right" value is subjective.
In the model later it won't be 258 to 586 in the uv either.

I did ask Razz, she said both version make sense, and with it being such an easy change for the user I don't think this really matters much.

JannisX11:
Alright, changed it to the native resolution. Looks good?

PR #119, opened 2020-12-27
Changed Importer to import the model name and id.

StealthyExpertX:
Changed Importer to import model name by importing the filename and setting the project name and geometry name and texture name.

### DIFF
--- a/plugins/csmodel.js
+++ b/plugins/csmodel.js
@@ -591,6 +591,10 @@ Plugin.register('csmodel', {
 					} else {
 						codec.parse(files[0].content);
 					}
+					csname = files[0].name.replace(".csmodel","").replace(" ","_").toLowerCase();
+					Project.name = csname;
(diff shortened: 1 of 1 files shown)


plugins/csmodel.js
```diff
@@ -591,6 +591,10 @@ Plugin.register('csmodel', {
 					} else {
 						codec.parse(files[0].content);
 					}
+					csname = files[0].name.replace(".csmodel","").replace(" ","_").toLowerCase();
```
JannisX11, line 594: Make sure to use regex to replace the white spaces with underscores. It's a little quirk in Javascript that if you replace a string, it will only replace the first occurrence.
e. g. `.replace(/\s+/, '_')`

StealthyExpertX: Gotcha and thanks for the pointer, I'll correct that.

PR #122, opened 2021-01-08
Added Double Sided Cubes Plugin

SnaveSutit:
Simply duplicates the selected Cube(s), flips them on all axis, inverts their scale and inflation, and fixes their position to line up with the original cube. This inversion causes a double sided rendering effect on java and I thought it would be nice to have a simple plugin that does all of these actions for you. Thanks to dragonmaster95#3249 for the solution to double sided rendering that doesn't require a plane per face of the cube.

### DIFF
--- a/plugins/double_sided_cubes.js
+++ b/plugins/double_sided_cubes.js
@@ -1,4 +1,37 @@
 (()=> {
+	const cube_action = new Action({
+		id:"create_double_sided_cubes",
+		name: 'Create Double Sided Cube',
+		icon: 'flip_to_back',
+		category: 'edit',
+		condition: () => !Project.box_uv,
+		click: function(ev) {
+			if (selected.length === 0) {
+				Blockbench.showMessage('No cubes selected', 'center')
+				return;
+			};
+			Undo.initEdit({elements:[]});
+			let cubes = [];
+			Cube.selected.forEach((cube) => {
+				const new_cube = new Cube({...cube}).init();
+				new_cube.name = new_cube.name + " inverted";
+				if (cube.parent !== "root") {
+					new_cube.addTo(cube.parent);
+				};
+				for (i=0; i<3; i++){
+					new_cube.flip(i, 0, false)
+				};
+				new_cube.from = [...cube.to];
+				new_cube.to = [...cube.from];
+				new_cube.inflate = -new_cube.inflate;
+				Canvas.adaptObjectPosition(new_cube);
+				Canvas.updateUV(new_cube);
+				cubes.push(new_cube);
+			});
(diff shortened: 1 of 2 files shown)


plugins/double_sided_cubes.js
```diff
@@ -0,0 +1,45 @@
...
+		    condition: () => !Project.box_uv,
+		    click: function(ev) {
+		    	if (selected.length === 0) {
+						Blockbench.showMessage('No cubes selected', 'center')
+						return;
+					}
+					Undo.initEdit({elements:[]});
+					let cubes = [];
+					Cube.selected.forEach((cube) => {
+						new_cube = new Cube({...cube}).init();
```
JannisX11, line 24: Declare a variable with one of the variable keywords (var, let, const)

plugins/double_sided_cubes.js
```diff
@@ -0,0 +1,45 @@
...
+						for (i=0; i<3; i++){
+							new_cube.flip(i, 0, false)
+						};
+						new_cube.from = [...cube.to];
+						new_cube.to = [...cube.from];
+						new_cube.inflate = new_cube.inflate * -1;
+						cubes.push(new_cube);
+					});
+					Undo.finishEdit('create_double_sided_cube', {elements:cubes});
+					Canvas.updateAll();
```
JannisX11, line 36: The use of `Canvas.updateAll()` is discouraged because of it's performance impact.

Since you are only updating the position and UV mapping of the new cube, use
* `Canvas.adaptObjectPosition(cube)`
* `Canvas.updateUV(cube)`

plugins/double_sided_cubes.js
```diff
@@ -0,0 +1,45 @@
+(()=> {
+	Plugin.register('double_sided_cubes', {
+		title: 'Double Sided Cubes',
+		author: 'SnaveSutit',
+		description: 'Creates inverted duplicates of the selected cube(s) to allow double-sided rendering in java edition.',
+		icon: 'flip_to_back',
+		version: '1.0.0',
+		variant: 'both',
+		onload() {
+			cube_action = new Action({
```
JannisX11, line 10: Declare a variable with one of the variable keywords (var, let, const)

In this case, you should declare it above `Plugin.register` so that it's available in the `onunload` event as well.

JannisX11 (commented):
Added some comments, mostly about best practices.

SnaveSutit:
Thanks for pointing those out, I'm still somewhat new to JS so I've not really got it fully down yet.

JannisX11:
Sorry I wasn't super clear about where the action should be defined. Check out how other plugins such as "Plaster" do it.

SnaveSutit:
That better?

JannisX11:
Looks good

PR #125, opened 2021-01-26
Added Bedrock Block Exporter

Brennian:
Created new plugin for creating new Blocks on Bedrock Edition

### DIFF
--- a/plugins/arcaniax_block_exporter.js
+++ b/plugins/arcaniax_block_exporter.js
@@ -1,0 +1,450 @@
+(function() {
+	
+	var geometry_name;
+	var geometry_name_fancy;
+	var bbMinSize = 0;
+	var bbMaxOffset = 8;
+	var bbMaxSize = 16;
+	var color;
+	var folderLocation;
+	let options = {};
+	let collision = [16, 16, 16];
+	let collisionOffset = [0, 0, 0];
+	
+	const fs = require('fs');
+	
+	var preview;
+	
+	
+	// SETS UP THE PREVIEW FOR THE COLLISION
+	function setupPreview(){
+		
+		if (preview!=null){
+			scene.remove(preview);
+		}
+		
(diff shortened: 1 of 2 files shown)


plugins/arcaniax_block_exporter.js
```diff
@@ -0,0 +1,450 @@
...
+		title: 'Bedrock Block Exporter',
+		icon: 'icon-format_block',
+		author: 'Arcaniax',
+		description: 'Helps making new Bedrock blocks (requires experimental mode)',
+		version: '0.1.1',
+		variant: 'both',
+
+		onload() {
+
+			Blockbench.addMenuEntry('Bedrock Block Exporter', 'icon-format_block', function() {
```
JannisX11, line 437: Please use Actions here. You can find an example in many newer plugins, e. g. "Plaster".
`addMenuEntry` is a deprecated method from the old days of Blockbench.

plugins/arcaniax_block_exporter.js
```diff
@@ -0,0 +1,450 @@
...
+	}
+	
+	
+	function getResourceManifest(formData){
+		var name = formData.name;
+		var uuid1 = generateUUID();
+		var uuid2 = generateUUID();
+		var json =
+`{
+    "format_version": 2,
```
JannisX11, line 260: Can't help but notice that the JSON files all use a different type of indentation. Maybe use a consistent method of indentation, like only tabs.
You could also use the built-in function `compileJSON()` to generate the JSON strings, then you won't need to worry about indentation and you can also write the JSON code directly in code, without having to deal with template strings.

JannisX11:
Also, please add the plugin to the `plugins.json` file.

plugins/arcaniax_block_exporter.js
Brennian: Changed to the newer method.

PR #128, opened 2021-01-27
Several Fixes in Mod Utils Plugin

### LINKED ISSUE #118: [Suggestion] Make voxel shape exports more intuitive
Within the UI it is not explained that elements used in a voxel shape export need to be grouped in a group named "VoxelShapes". Also, the resulting Java code may be confusing to new developers, as it does not create a `VoxelShape` but an `Optional<VoxelShape>`. I would suggest the following stream reduction format instead:
```java
// Take an empty shape as an identity
// Use the method that delegates to combineAndSimplify(a, b, OR)
// Return a VoxelShape rather than an Optional<VoxelShape>
.reduce(VoxelShapes.empty(), VoxelShapes::or);
```
Alternatively, Mojang already have a static helper that does this too!
```java
// Essentially Arrays.stream(varargs).reduce(firstArg, Shapes::or)
VoxelShapes.or(VoxelShapes.empty(), Block.createCuboidShape(...), Block.createC...);
```
(Fwiw this is how I discovered how to export a voxel shape when I had to Google it... Definitely needs to be clearer: https://www.reddit.com/r/Blockbench/comments/k3omsn/voxelshape_wont_export/?utm_source=ifttt)

### LINKED ISSUE #123: [Mod Utils] Tabula imported models with mirroring not mirroring correctly
How it should look: 
image

After imported into Blockbench via Mod Utils.
image

Test Mode: 
[ICON_OF_SIN.zip](https://github.com/JannisX11/blockbench-plugins/files/5828752/ICON_OF_SIN.zip)

JTK222:
Fixes #123
Partially fixes #118


JannisX11:
Totally forgot to merge this :(

PR #140, opened 2021-03-20
Added simplify plugin

RyanGar46:
Simplifies the cubes in a model. For example, if a block was 0.99 pixels wide, then it would change it to 1.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -309,5 +309,13 @@
 		"version": "2.0.0",
 		"min_version": "3.0.0",
 		"variant": "both"
+	},
+	"simplify": {
+		"title": "Simplify Models",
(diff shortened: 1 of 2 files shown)


JannisX11:
Three small issues:
- No Undo functionality. Users won't be able to undo what the plugin did. Check the documentation and other plugins such as Plaster for reference on how to make modifications undo compatible.
- Filter affects all elements, but makes modifications that would only work on cubes. Please run the filter only on cubes (`Cube.all`) and not on all elements.
- Description says that it affects the scale of cubes. This should probably say size instead.

RyanGar46:
That should have fixed the issues.

PR #150, opened 2021-04-20
Add JavaAnimator [Plugin]

Blocksnmore:
JavaAnimator is a plugin where you can create models and have them exported as models for a resource pack which could be used for easy animating.


DoubleF3lix:
FYI, Blockbench 4.0 changed `ModelMeta.export_path` to `Project.export_path`.

PR #161, opened 2021-07-10
Adding Array Tool Plugin[Now Fixed]

Malik12tree:
A Plugin that adds an Array tool that creates an array of copies of the base object, with each copy being offset from the previous one in any of a number of possible ways.
Screenshot from 2021-07-10 03-11-02

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -323,6 +323,17 @@
 		"version": "0.2.0",
 		"variant": "both"
 	},
+	"animation_sliders": {
+		"title": "Animation Sliders",
+		"icon": "fas.fa-bezier-curve",
+		"author": "JannisX11",
+		"description": "Adds multiple sliders to tweak keyframes",
+		"about": "Adds sliders and other tools to modify keyframes:\n\n- **Tween Keyframes:** Amplify the values of the selected keyframes\n- **Amplify Keyframes:** Amplify the values of the selected keyframes\n- **Ease Keyframes:** Create a curve with the selected keyframes between the adjacent keyframes\n- **Keyframe Slider Axis:** Select which axis the keyframe sliders affect\n- **Create Keyframe Column:** Key all channels in the timeline at the current timecode, if they already have keyframes\n- **Select Keyframe Column:** Select all keyframes in the timeline along a column below the playhead",
+		"tags": ["Animation"],
+		"version": "0.1.0",
+		"min_version": "3.7.0",
+		"variant": "both"
+	},
 	"array_tool": {
 		"title": "Array tool",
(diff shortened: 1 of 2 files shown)


JannisX11:
I found a couple of issues with the plugin. Please fix those before the plugin can be accepted into the plugin store.

1. The code only operates on cubes, but you are targeting all selected elements. Run the function on `Cube.selected` instead of `selected`.
2. You are creating multiple undo points, and some of them overlap each other. You should only be creating one undo point.
3. Undo points keep track of all elements, not just the ones created. On large models this can be bad for performance. Instead, create an empty array, initiallize the undo point with it, add the new elements to the array, then finish the undo point.
4. Created cubes appear to have inverted dimensions
5. Created cubes appear at the bottom of the outliner, rather than next to the original

This is the default code to duplicate elements in Blockbench. You can use the default `duplicate()` method on cubes to avoid the last issue and to support all properties on cubes. The code sample also shows how you can solve issue 2 and 3.

```javascript
	var added_elements = [];
	Undo.initEdit({elements: added_elements, outliner: true, selection: true})
	selected.forEachReverse(function(obj, i) {
		var copy = obj.duplicate();
		added_elements.push(copy);
	})
	BarItems.move_tool.select();
	Undo.finishEdit('Duplicate elements')
```

Malik12tree:
> I found a couple of issues with the plugin. Please fix those before the plugin can be accepted into the plugin store.
> 
> 1. The code only operates on cubes, but you are targeting all selected elements. Run the function on `Cube.selected` instead of `selected`.
> 2. You are creating multiple undo points, and some of them overlap each other. You should only be creating one undo point.
> 3. Undo points keep track of all elements, not just the ones created. On large models this can be bad for performance. Instead, create an empty array, initiallize the undo point with it, add the new elements to the array, then finish the undo point.
> 4. Created cubes appear to have inverted dimensions
> 5. Created cubes appear at the bottom of the outliner, rather than next to the original
> 
> This is the default code to duplicate elements in Blockbench. You can use the default `duplicate()` method on cubes to avoid the last issue and to support all properties on cubes. The code sample also shows how you can solve issue 2 and 3.
> 
> ```js
> 	var added_elements = [];
> 	Undo.initEdit({elements: added_elements, outliner: true, selection: true})
> 	selected.forEachReverse(function(obj, i) {
> 		var copy = obj.duplicate();
> 		added_elements.push(copy);
> 	})
> 	BarItems.move_tool.select();
> 	Undo.finishEdit('Duplicate elements')
> ```

so i should make the for loop run this function or u mean something else?

Malik12tree:
I fixed every issue now.

PR #166, opened 2021-07-21
Array Tool Plugin, Aborted(dont want to submit anymore)

Malik12tree:
A Plugin that adds an Array tool that creates an array of copies of the base object, with each copy being offset from the previous one in any of a number of possible ways.


JannisX11:
Please fix the issues I mentioned in #161 before the plugin can get accepted.

At the moment, the plugin still throws an error when anything is selected that is not a cube. Cubes don't end up next to the original so parent group rotation is not applied. Properties like UV mapping when using Box UV are not copied.

Also, it's easier to keep track of the progress if you keep this PR and don't open a new PR for every change.

Malik12tree:
@JannisX11  The Plugin has been finnaly fixxed from every issue, Thanks.

PR #170, opened 2021-08-10
Image Model Importer Plugin

Malik12tree:
import an image to Turn it into a 3d model.
Screenshot from 2021-08-10 07-10-11


JannisX11:
This plugin seems to be duplicating default Blockbench functionality by doing essentially the same thing as importing an Extruded Texture using "Pixel" scan mode.

Malik12tree:
> This plugin seems to be duplicating default Blockbench functionality by doing essentially the same thing as importing an Extruded Texture using "Pixel" scan mode.

I didnt know that exists...

PR #172, opened 2021-08-15
Add Armor Stand Animator Plugin

DoubleF3lix:


JannisX11:
Please use the tag `Minecraft: Java Edition`. I've also added some informations about this in README.MD.

Also, it might be helpful to include a quick usage instruction in the about field.

DoubleF3lix:
I've made the changes per your suggestions.

PR #175, opened 2021-09-13
[Geckolib Plugin] Fixes Swapped Arms Names

AzureDoom:
This pr just fixes the Right and Left arm names being swapped in new Geckolib Armors.


JannisX11:
I think you need to recompile the actual plugin

AzureDoom:
Alright, the plugin is updated properly this time! I've also bumped up the max version to support the new 4.x releases as it is working from my testing.

PR #192, opened 2021-11-14
Vintage Story Models

Malik12tree:
Allows exporting Vintage Story models

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -382,11 +382,11 @@
 		"variant": "desktop"
 	},
 	"vintagestory_models": {
-		"title": "Reference Models",
-		"author": "Vintage Story Models",
+		"title": "Vintage Story Models",
+		"author": "Malik12tree",
 		"icon": "park",
 		"description": "Allows to export VintageStory models.",
-		"version": "1.0",
+		"version": "1.0.0",
 		"variant": "both"
 	}
 }
--- a/plugins.json
+++ b/plugins.json
@@ -380,5 +380,13 @@
 		"version": "1.0.0",
 		"min_version": "4.0.0",
 		"variant": "desktop"
+	},
+	"vintagestory_models": {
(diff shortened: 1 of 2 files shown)


JannisX11:
Please adjust the plugin meta data. The plugin name is wrong, and the version should be semver (e. g. 1.0.0).

plugins/vintagestory_models.js
```diff
@@ -0,0 +1,249 @@
...
+            })
+            MenuBar.addAction(exportVsAction, "file.export");
+        },
+        onunload() {
+            exportVsAction.delete();
+        }
+    });
+})()
+
+function getRangeBool(x, min, max) {
```
IanSSenne, line 221: these should prob be moved into the IIFE as to not crowd the global scope.

plugins.json
```diff
@@ -380,5 +380,13 @@
 		"version": "1.0.0",
 		"min_version": "4.0.0",
 		"variant": "desktop"
+	},
+	"vintagestory_models": {
+		"title": "Reference Models",
+		"author": "Vintage Story Models",
+		"icon": "park",
+		"description": "Allows to export VintageStory models.",
+		"version": "1.0",
```
AzureDoom, line 389: You should be following https://semver.org/#summary so MAJOR.MINOR.FIX EX: 1.0.0

Malik12tree: ohh i didnt notice that, Thanks!!

JannisX11:
Please test this on some models. I noticed an issue on a simple model I tried to export. For example, in line 46 you iterate over groups, but in the same function in 57 you are referring to cubes, which can cause a type error if there are more Groups than cubes.

Malik12tree:
oh ok, I nearly tried on 4-5 different models and fixed some stuff, I think  **Now** everything should be done

PR #194, opened 2021-11-27
Add Endimator Animations Exporter Plugin

SmellyModder:
Adds a plugin for exporting Blockbench animations as JSON files in the format of _Endimator_ Animations (Endimations).

For context, _Endimator_ is a sub-library in the [Blueprint](https://github.com/team-abnormals/blueprint) Library Mod designed for complex keyframe animations on _Minecraft: Java Edition_ models. Animation JSON files used by _Endimator_ are in a unique format, and this plugin paves the way for animators making mods with _Endimator_ using Blockbench.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -388,5 +388,16 @@
 		"description": "Allows to export VintageStory models.",
 		"version": "1.0.0",
 		"variant": "both"
+	},
+	"endimations_exporter": {
+		"title": "Endimator Animations Exporter",
+		"author": "SmellyModder",
+		"icon": "icon-format_java",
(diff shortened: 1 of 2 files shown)


JannisX11:
I noticed when testing that the exporter assumes that all animators in an animation are BoneAnimators. This causes the export to fail if there are effect keyframes. Even if effect keyframes are not intended to be used, you should still fix this by checking if the animator is an instance of BoneAnimator before calling `animator.getGroup()`.

SmellyModder:
> I noticed when testing that the exporter assumes that all animators in an animation are BoneAnimators. This causes the export to fail if there are effect keyframes. Even if effect keyframes are not intended to be used, you should still fix this by checking if the animator is an instance of BoneAnimator before calling `animator.getGroup()`.

Ah yes, I thought about this before but totally forgot about it.
Fixed now 👍

PR #201, opened 2021-12-25
Add Animation Speed Changer [Plugin]

Ancientkingg:
There was really no way to accurately modify the keyframes themselves in an animation according to a different speed. So this plugin was made in response.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1,4 +1,13 @@
 {
+	"change_animation_speed": {
+        "title": "Animation Speed Changer",
+        "author": "Ancientkingg",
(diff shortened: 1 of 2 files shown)


JannisX11:
This is duplicating base Blockbench functionality. You can hold Ctrl to temporally stretch the selected keyframes in the timeline. Also, you can type an expression like `* 2` into the keyframe time field to multiply all their time values by two and double the length of the animation. Plus, for Minecraft Bedrock Edition, you can type something like `q.anim_time + q.delta_time * 2` to change the playback speed of the animation.

PR #202, opened 2021-12-30
Texture Stitcher: Implement mesh UV stitching

mchorse:
This PR makes texture stitching process work also with meshes. Thanks to wacky for reporting! 👍


JannisX11:
Don't forget to change the version number in plugins.json

PR #206, opened 2022-01-08
Add VoxelShape Fabric Generator

CyberStefNef:
Generate Voxel Shape files for Fabric


CyberStefNef:
I have changed my plugin script. I put the functions into the main function. That should be fine.

PR #227, opened 2022-04-24
 MTools 1.0.1 update

Malik12tree:


Trioplane:
epic plugin

plugins/mesh_tools.js
```diff
@@ -1715,6 +1806,421 @@
...
+			quickPrimitives.push(
+				new Action("polyhedron", {
+					name: "Polyhedron",
+					icon: quickPrimitivesI,
+					click(){
+						new Dialog ({
+							title: "Quick Primtive [ Polyhedron ]",
+							form: {
+								select: {label: "Hedron", type:"select", options: {Icosahedron:"Icosahedron", Dodecahedron:"Dodecahedron",Octahedron:"Octahedron", Tetrahedron:"Tetrahedron"}},
+								radius: {label: "Radius", value: 8},
```
JannisX11, line 2140: These two lines should be of type "number", especially because they are used in eval. Using plain strings from user inputs would be a security risk.

plugins/mesh_tools.js
```diff
@@ -1715,6 +1806,421 @@
...
+			);
+			let quickPrimitives = [];
+			let quickPrimitivesI = "offline_bolt";
+			quickPrimitives.push(
+				new Action("polyhedron", {
+					name: "Polyhedron",
+					icon: quickPrimitivesI,
+					click(){
+						new Dialog ({
+							title: "Quick Primtive [ Polyhedron ]",
```
JannisX11, line 2137: Noticed a typo, it should be "Primitive".

plugins/mesh_tools.js
```diff
@@ -1715,6 +1806,421 @@
...
+					}
+				})
+			);
+			quickPrimitives.push(
+				new Action("torusknot", {
+					name: "Torus Knot",
+					icon: quickPrimitivesI,
+					click(){
+						new Dialog ({
+							title: "Quick Primtive [ Torus Knot ]",
```
JannisX11, line 2175: Same typo as above

JannisX11 (commented):
Found 3 little things.

plugins/mesh_tools.js
JannisX11: I'd generally avoid eval if possible. In this case, you can do the same thing without it.
```
new THREE[out.select + 'BufferGeometry'](s.radius, s.detail-1)
```

Malik12tree:
Sorted!

PR #233, opened 2022-05-09
Fix update selection bug

ewanhowell5195:
as the title says


JannisX11:
You need to check for the format id, not the name. Otherwise it's only gonna work in English.

PR #236, opened 2022-05-18
Adding BoxUV Cube Flagger Plugin

SirJain0:
This plugin flags all cubes less than one unit on any axis. Users can flag these invalid cubes with one click (Tools > Flag Invalid Cubes) and can edit them so they are no longer invalid. This prevents the step of having to convert to Per-Face UV if they don't want to, or if it is not an option in the format.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -463,13 +463,13 @@
 		"variant": "both"
 	},
 	"boxuv_cube_flagger": {
-		"title": 'BoxUV Cube Flagger',
-		"author": 'SirJain',
-		"description": 'This plugin flags cubes less than 1 unit by flashing them.',
-		"icon": 'lightbulb',
-		"version": '1.0.0',
-		"variant": 'both',
-		"min_version": '4.2.0',
+		"title": "BoxUV Cube Flagger",
+		"author": "SirJain",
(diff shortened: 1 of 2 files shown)


JannisX11:
I still noticed some issues while using the plugin:
- If I use the action with no "invalid" cubes in the scene, and then add one while it would still be flashing, I get console errors and the screen freezes.
- If I click the action again while some elements are red, they just stay red forever.

Also, cubes with a width of exactly 0 (including inflation) are not an issue generally, something to consider.

SirJain0:
I'll try to fix these, should I open an entirely new PR for the updated version or what's the process for submitting the same plugin but with fixed bugs?

JannisX11:
You should add commits to this PR while it's still open. Once it's merged, and you want to submit a new version, you can create a new PR.

SirJain0:
I fixed up the plugin and I edited the file directly. The following bugs have been fixed:
- Screen freezing when new invalid element is added during flashing (they are now ignored for that flashing)
- Fixed the continuous red when the action is clicked while it's already running (the action now gets ignored)
- The plugin now ignores cubes that are 0 pixels on an axis, as these are not actually invalid
- Fixed a bug where the plugin would stop working when a flagged cube is deleted while the flagging is in progress

PR #239, opened 2022-05-25
Tint Preview Plugin

MrCrayfish:
Tint Preview is a tool to help designers who work with the tint face feature of Java Block/Item models. It features the ability to quickly preview a color on tint enabled faces. Tint Preview adds two new actions to the Tools menu that allows you to toggle tinting and change the tint color.

image

This is my first plugin for Blockbench, any feedback is appreciated and hope it gets merged.


JannisX11:
Looks nice!

I have a few points of feedback:

* When testing, I sometimes had issues with the preview not updating. Disabling tint preview would not disable it, and while disabled, changing the color would not update the color in the preview.
* For the toggle, instead of using an Action, you can use a new Toggle(). It works similarly, but has the advantage that it shows the status in menus and toolbars.
* Regarding translations, yes, they are correctly implemented, although they don't do much currently, since it's only English, so you could have just used the names directly. But if you plan to add more languages in the future, it's helpful.
* Also, as a suggestion for a potential future update: You could add some simple presets for biome foliage colors, since that's the most common use case for resource pack creators.

Will merge as soon as the updating issue is resolved 👍

JannisX11:
Oh, one more thing: There is an error message when moving elements that are not cubes, such as locators: (these don't have a "faces" field).
image

MrCrayfish:
- Fixed issue with tinting not updating correctly
- Switched to toggle for "show tint" action

MrCrayfish:
I'll have a patch coming soon. I discovered a minor issue when first installing the plugin that tinting doesn't work for opened projects. This is only on the initial install, reopening the project or restarting Blockbench fixes the issue. I'll probably take the time to also include presets for grass, leaf colours, etc.

PR #240, opened 2022-05-29
Add OptiFine Player Models and update CEM Template Loader

ewanhowell5195:
The OptiFine Player Models plugin is a plugin that adds a new format to allow you to create player models to use with OptiFine.

This pull request also updates CEM Template Loader to fix some bugs.


JannisX11:
When styling dialog content, make sure to scope it to the dialog by adding `dialog#<dialog id>` in front of the CSS selector, or by using inline styles. For example, when opening the about window, you can see that the styling of the outliner changes.

PR #245, opened 2022-06-02
Add quick_box_uv_layout

Manuel-3:
This plugin adds a button that rearranges the UV of selected cubes in Per-face mode into the shape of Box-UV mode.

https://user-images.githubusercontent.com/58179488/171749398-531b4f82-9fe3-407a-a5cb-adc6c6b259e5.mp4

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -494,5 +494,14 @@
 		"min_version": "4.2.0",
 		"tags": ["Minecraft: Java Edition"],
 		"about": "Tint Preview is a tool to help designers who work with the tint face feature of Java Block/Item models. It features the ability to quickly preview a color on tint enabled faces. Tint Preview adds two new actions to the Tools menu that allows you to toggle tinting and change the tint color.\n**Important:** This plugin is designed for Java Block/Item models only and will not work for other formats."
+	},
+	"quick_box_uv_layout": {
+		"title": "Quick Box-UV Layout",
+        "author": "Manuel-3",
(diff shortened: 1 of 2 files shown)


JannisX11:
Nice idea!
If the intention is to generate box UV compatible UV maps, the top and bottom faces should be flipped. The up face on both axes, the down face only on the vertical axis.
Also a nice QoL improvement would be to generate the UV at the current offset, so that if you press the button again it doesn't shift the UV into the top left corner.

JannisX11:
Also please make sure to copy the value of the "about" field into the plugin file too.

Manuel-3:
@JannisX11 Thanks for your feedback! I really didn't think of flipping the top and bottom faces.

This is now fixed, also added the QoL thing for if the user is in box uv and then clicks the button, it will not reset it to the top left corner. (I think this is what you meant? For Per-face there is no way to get a proper offset, apart from doing some kind of average or something as far as I can tell?)

JannisX11:
Yeah I meant in per-face UV. You can either find the top left corner of all faces, by calculating the smalles value of all UV coordinates of all faces. Or you can calculate the center, which might be more difficult.

Manuel-3:
@JannisX11 Ah ok I see what you mean. It should stay in the same place when clicking the button multiple times, instead of moving it to top left. Added that, should be ready to go now.

JannisX11:
Excellent!

PR #246, opened 2022-06-02
Pie Menus

Malik12tree:
Add pie shaped menus for faster work.


JannisX11:
Looks really nice! I've always wanted to make a pie menu plugin, but never got around to it.
Two notes:
* The plugin currently breaks the search functionality inside the keybinding menu and in action control. I think this is partly because the menus you are creating have no ID.
* Add "Pie Menu" to the names of your pie menus so that they are easier to find in the keybindings list.

PR #248, opened 2022-06-08
Update "modded_entity_fabric.js" to include 1.17+ entity changes

SebaSphere:


plugins/modded_entity_fabric.js
```diff
@@ -176,7 +260,8 @@ Plugin.register('modded_entity_fabric', {
 	},
 	onunload() {
 		delete Codecs.modded_entity.templates['Fabric 1.14'];
-		delete Codecs.modded_entity.templates['Fabric 1.15+'];
+		delete Codecs.modded_entity.templates['Fabric 1.15-1.16'];
```
fadookie, line 263: I think here you may want to do:
```js
		delete Codecs.modded_entity.templates['Fabric 1.15+'];
		delete Codecs.modded_entity.templates['Fabric 1.15-1.16'];
```

because in the case they are upgrading this plugin it's going to hot-swap it, you want to remove the old template added by the old version of the plugin.

fadookie:
Thanks for adding 1.17 support for fabric! Had a few questions.

plugins/modded_entity_fabric.js
SebaSphere: Updated this in a new commit.

Also added a comment in the code mentioning where the code was inspired from should someone else be utterly confused about what's happening 😂

plugins/modded_entity_fabric.js
```diff
@@ -81,14 +81,13 @@ function setTemplate() {
 		renderer: `%(bone).render(f5);`,
 		cube: `%(bone).boxes.add(new Box(%(bone), %(uv_x), %(uv_y), %(x), %(y), %(z), %(dx), %(dy), %(dz), %(inflate), %(mirror)));`,
 	};
-
-	Codecs.modded_entity.templates['Fabric 1.15+'] = {
-		name: 'Fabric 1.15+',
+	Codecs.modded_entity.templates['Fabric 1.15-1.16'] = {
```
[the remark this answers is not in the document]
JannisX11: Yes, it would fall back to the latest Forge version if the version that the project was saved in no longer exists.
I would recommend instead to only change the name, not the ID.

JannisX11:
Looks good to me! @fadookie permission to merge?

fadookie:
@JannisX11 thanks, LGTM! 👍

PR #253, opened 2022-06-18
Add plugin to convert animation to java code

MatthewGreenspun:
Has been tested and confirmed to be working on Forge and Fabric for 1.19.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -532,6 +532,7 @@
 		"icon": "fa-cube",
 		"version": "1.0.0",
 		"variant": "both",
-		"about": "This plugin exports your blockbench animations as java code to be used for the new 1.19 keyframe system. Please note that this system does not support Molang or step interpolation"
+		"about": "This plugin exports your blockbench animations as java code to be used for the new 1.19 keyframe system. Please note that this system does not support Molang or step interpolation",
+		"tags": ["Minecraft: Java Edition"]
 	}
 }
--- a/plugins.json
+++ b/plugins.json
@@ -524,5 +524,14 @@
 		"version": "0.0.1",
 		"variant": "both",
 		"about": "This plugin simply adds a button that rearranges your individual Per-face UV rectangles in the same way Box-UV mode would. This is handy if you want to use the standard Box-UV format, but also have the flexibility of Per-face UV mode."
+	},
+	"animation_to_java": {
+		"title": "Animation to Java Converter",
+		"author": "MG",
+		"description": "Converts Blockbench animations to Java code for the new 1.19 keyframe system",
+		"icon": "fa-cube",
+		"version": "1.0.0",
+		"variant": "both",
+		"about": "This plugin exports your blockbench animations as java code to be used for the new 1.19 keyframe system. Please note that this system does not support Molang or step interpolation"
 	}
 }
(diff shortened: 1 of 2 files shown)


plugins.json
```diff
@@ -524,5 +524,14 @@
...
 		"variant": "both",
 		"about": "This plugin simply adds a button that rearranges your individual Per-face UV rectangles in the same way Box-UV mode would. This is handy if you want to use the standard Box-UV format, but also have the flexibility of Per-face UV mode."
+	},
+	"animation_to_java": {
+		"title": "Animation to Java Converter",
+		"author": "MG",
+		"description": "Converts Blockbench animations to Java code for the new 1.19 keyframe system",
+		"icon": "fa-cube",
+		"version": "1.0.0",
+		"variant": "both",
```
JannisX11, line 534: Meta data must be the same in plugins.json and in the plugin itself. Here the tags are missing.

plugins/animation_to_java.js
```diff
@@ -0,0 +1,84 @@
...
+  var menuButton;
+
+  Plugin.register("animation_to_java", {
+    title: "Animation to Java Converter",
+    author: "MG",
+    description:
+      "Converts Blockbench animations to Java code for the new 1.19 keyframe system",
+    icon: "fa-cube",
+    version: "1.0.0",
+    variant: "both",
```
JannisX11, line 11: The about text is missing here.

plugins/animation_to_java.js
```diff
@@ -0,0 +1,84 @@
...
+    tags: ["Minecraft: Java Edition"],
+    onload() {
+      menuButton = new Action("export_to_java", {
```
JannisX11, line 14: The ID should include animation. IDs are global so they should be specific to reduce the risk of conflict.

plugins/animation_to_java.js
```diff
@@ -0,0 +1,84 @@
...
+        name: "Export Animations to Java",
+        description: "Exports animations to Java code",
+        icon: "fa-file-export",
```
JannisX11, line 17: I would recommend to add a condition here so that the action doesn't show up in all formats.
Something like:
`condition: () => Format.animation_mode`

plugins/animation_to_java.js
```diff
@@ -0,0 +1,84 @@
...
+        click: function () {
+          var file = generateFile();
+          Blockbench.export({
+            type: "Text File (.txt)",
```
JannisX11, line 21: No need to include txt here, that's automatically added
image

JannisX11 (commented):
Just a few notes:

plugins/animation_to_java.js
JannisX11: Add the condition for the action, not for the entire plugin. Otherwise it will also not work when switching to formats with animation mode later.

plugins/animation_to_java.js
```diff
@@ -0,0 +1,87 @@
...
+      .replaceAll(".", "_")
+      .replace("animation_", "")
+      .toUpperCase()} = AnimationDefinition.Builder.withLength(${
+      animation.length
+    }f)`;
+    if (animation.loop === "loop") {
+      outfileText += ".looping()";
+    }
+    for (const id in animation.animators) {
+      const boneAnimator = animation.animators[id];
```
JannisX11, line 52: Add a check here to make sure the plugin doesn't fail if non-bone animators exist:
`if (boneAnimator instanceof BoneAnimator == false) continue;`

PR #257, opened 2022-07-02
added Tabula Exporter

grillo78:


grillo78:
a friend reported a bug just after opened the pull request

PR #261, opened 2022-07-14
Updated tag to be less confusing.

dragonmaster95:
Old tag was confusing people into thinking it works in all Minecraft related formats, not just in the Java Block/Item one.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -115,7 +115,7 @@
 		"title": "Player Statue Generator",
 		"author": "Wither, dragonmaster95 and 3XH6R",
 		"description": "Generates player shaped models.",
-		"tags": ["Minecraft"],
+		"tags": ["Minecraft: Java Edition"],
 		"icon": "accessibility",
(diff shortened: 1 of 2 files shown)


dragonmaster95:
Actually, do we have a tag for only the Java Block/Item format? (since it literally only works for that format)

SirJain0:
You should update plugins.json too, right? (Let me know if I'm wrong)

JannisX11:
Yes, please update plugins.json too.
However, I think there is no reason this should only work in Java Block / Item, right? Just remove the condition, add a check before setting display settings, and let it work in all formats.

dragonmaster95:
I didn't make the plugin, only the model 🤷‍♂️ 
Was just annoyed by people constantly asking how they use the plugin in other formats cause just the tag "Minecraft" made it sound like it works in all Minecraft intended formats.

JannisX11:
Just to be clear, the plugins.json file has to be updated too, otherwise this won't work.

PR #268, opened 2022-07-29
Added root motion extractor plugin

Tschipp:
Adds a plugin which tries its best to estimate a velocity for an animation, and then gives you the values needed to reach that velocity for the movement component.
image
image


JannisX11:
That's kinda impressive!

PR #278, opened 2022-09-16
Add Collapsible Start Screen Categories plugin

ewanhowell5195:
This plugin adds a small dropdown arrow next to each category on the start menu, allowing you to collapse each category and hide their respective formats.


JannisX11:
Menu is fairly ambiguous, you should call it Start Screen or Formats, ideally directly in the plugin name.

PR #282, opened 2022-09-26
Add Repeat It plugin

Ocraftyone:
Plugin is similar to Blender's array feature where it modifies a single cube by a user determined amount iteratively.


JannisX11:
Feedback:
* `plugins.json` has incorrect indentation, please make sure it uses tabs and is consistent with the rest of the file
* It does not seem to work in formats with limited rotation. Even when the rotation offset is set to 0, it shows the error message box.
* Cube faces are not copied. In order to clone an entire element, you can pass the original element as the first argument in the constructor. Then you can use `.extend({})` to set your custom values.
* Since this only works for cubes, could you either add that to the description (probably easier), or make sure it works for all default element types?
* Leaving the repeat number at 1 does nothing, so I would recommend to either have "1" make one copy, or set the default value to 2.

Ocraftyone:
Thank you. I will work on that when I get some free time

Ocraftyone:
Ok. I think that I got everything stated plus a couple others I came across. Have another look over it.

Ocraftyone:
Merged all of the changes to make sure there are no issues in the pr. Should be good to go still

JannisX11:
Cool. One more thing: New elements are always added to the root. It would make more sense to add them to the parent of the original cube, since otherwise the rotations of the parent groups and the organization gets lost.
Also, could you change the description from Shapes to Cubes. Since most people don't read the "about" section. It would be important to make that clear.

Ocraftyone:
@JannisX11 Fixed that

JannisX11:
Did you test this with rotated groups? You'll need to update the position of the cubes after adding them to the group. You can do this via `Canvas.updateView({ elements: [cube], element_aspects: {transform: true} })`.
Also, you missed the second point in my previous comment.

Ocraftyone:
@JannisX11 not exactly sure if I implemented that statement correctly, but I have an implementation. If not, I went ahead and allowed you to make changes that way you can make changes if it is something simple. I also updated the desc.

PR #287, opened 2022-10-09
Add Custom Marker Colors plugin

SirJain0:
This plugin adds custom marker colors which are compatible with meshes and keyframes. A couple things are planned for this plugin in the future, but the plugin works well.

image
image


SirJain0:
I want to add more things to this plugin actually, so I'll close this for now.

PR #290, opened 2022-10-16
Add Custom Markers Plugin

SirJain0:
Allows users to add marker colors which can also be used for meshes and keyframes.
Manage marker colors (view or delete) via `Marker Color Menu -> Edit Markers` and add new ones using `Marker Color Menu -> Add Custom Marker`.
Marker IDs are generated based on the marker name, but lower-cased and with underscores instead of spaces.
For more info, you can check out the about dialog via `Help -> About Plugins -> About Custom Marker Colors...`


JannisX11:
* In my test, colors did not persist after restarting Blockbench
* Why did you call it Marker, not Marker Color in some places?

SirJain0:
Just fixed some merging conflicts, and I'll take a look at those points.

SirJain0:
I fixed all the above. Please take a look.

JannisX11:
Cool, that works now!
I also noticed that the options to edit marker colors only appear in the cube context menu. They are missing on groups, meshes, keyframes, and timeline markers. I think in order to avoid confusion and to future-proof the plugin for color-codable features in future Blockbench updates, it would be good to add the options to the main menu bar. Maybe as a sub-menu under File > Preferences..., or in the Tools menu.

SirJain0:
@JannisX11 - I made those changes, you can now find the buttons in a submenu inside the tools menu. Works a lot better.

PR #292, opened 2022-10-17
Plugin Loader 2.0

ewanhowell5195:
This plugin replaces the built in Blockbench Plugin Loader with a new and improved Plugin Loader.

## Plugin Page
Each plugin now gets its own page where you can view all of its information, as well as install/uninstall/reload it.

## Plugin Tabs
All of the tags used across the plugins are now shown as a list of tabs up the side of the Plugin Loader, allowing you to quickly and easily browse and filter by the tags.


JannisX11:
Just one thing: Could you clarify in the description that this changes the interface, but not the plugin loader itself? Maybe replace Plugin Loader in the description with Plugin Browser or Plugin Dialog.

PR #296, opened 2022-10-25
Update "Animation to Java Converter" plugin

DailyCraft:
Update the plugin to activate the "Animate" tab on the "Modded Entity" format to allow the plugin to be used with the format for which it is made.


MatthewGreenspun:
Looks good to me!

PR #301, opened 2022-11-07
added rounding function to animation_to_java.js to fix formatting

VincentHuto:
Before this when exporting to a Java animation txt file, not only was each file line so long it would crash eclipse but it also had some weird line breaks meaning it couldn't be formatted or directly used without manual cleanup like this.
image
****
Ive added a round function to each float field that not only will reduce file size considerably but also fixes formatting so it can be directly copy and pasted without issue.
image


JannisX11:
I feel like this could work better with actual indentation, to distinguish keyframes from channels etc.
Also, could you update the version number in plugins.json as well when changing it in the plugin file?

Something unrelated to this PR that I noticed is that animators pull their name from the internal property _name, instead of .name. Which can result in the animation name being set as "undefined".

VincentHuto:
All very good points!, Ill take another pass over it and see what can be cleaned up, as for the _bone vs bone honestly not sure why the original author did it  ¯\_(ツ)_/¯

VincentHuto:
I will say even with proper indentation, this always seems to clip weird at this one scaleVec section without rounding for some odd reason
image

VincentHuto:
Indentation and plugin.json have been updated/cleaned up with the new commit

MatthewGreenspun:
Looks good to merge!
Honestly I don't mind if you stop including me in PRs. I originally created this plugin for a friend and he thought it would be a good idea to publish it but I don't even make mods myself and I trust the community to maintain the plugin much better than I could.

PR #305, opened 2022-11-14
Add Plugin Statistics Plugin

Malik12tree:


Malik12tree:
I also added a commit for the brush plus for the Brush Plus plugin:
- For the the App version: Use IndexDB to really save images into the memory instead of the path.

PR #310, opened 2022-12-06
Update Image Centering plugin to 1.1.0

SirJain0:
## Changelog
- Cleaned up code a bit
- Simplified some functions
- Centering can now be done with the keybind `c`


JannisX11:
Here's how you can define a keybinding: https://github.com/JannisX11/blockbench-plugins/blob/master/plugins/brush.js#L901
That way it's displayed in the tooltip and it can be re-assigned.

SirJain0:
@JannisX11 Thanks, that makes it a lot easier.

PR #311, opened 2022-12-14
Add Scene Recorder plugin

ewanhowell5195:
This plugin adds a new scene recorder that allows you to record your model and scene in a large variety of formats.

Using this plugin is very similar to using the built-in GIF Recorder.

You can find the `Record Scene...` button in the toolbar above the viewport.

## Formats
- **GIF**
  - Record animated GIFs in higher quality than the built-in GIF Recorder
- **MP4**
  - Record an MP4 video
- **WebM**
  - Record a WebM video
- **WebP**
  - Record an animated WebP image
- **APNG**
  - Record an animated PNG image
- **PNG Image Sequence**
  - Output each frame as an individual PNG image

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -270,6 +270,17 @@
 		"min_version": "4.4.3",
 		"variant": "both"
 	},
+	"scene_recorder": {
+		"title": "Scene Recorder",
+		"icon": "video_camera_back",
+		"author": "Ewan Howell",
+		"description": "Add a new scene recorder where you can record your model in a large variety of formats.",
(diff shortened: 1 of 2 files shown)


JannisX11:
Placing the action in the top toolbar is a bit inconsistent, as the toolbar by default is reserved for actual tools that are relevant to editing the model. You should place it in the View > Screenshot menu instead, that's where people who are familiar with the GIF recorder will look.

I had some issues with recording several of the formats, but they were all resolved by updating to the latest version of ffmpeg. The only issue that remains is that I am unable to open the recorded MP4 videos. I have tried with all codecs, and in different video players. The videos is attached.

https://user-images.githubusercontent.com/29278436/208271162-01e76ed1-bfb9-4aff-a759-5b5fcaa95976.mp4

PR #312, opened 2022-12-15
Image Centering bugfix 

SirJain0:
Fixes a bug where the installation quick message shows when Blockbench is opened, even if the plugin isn't being installed


JannisX11:
You can get rid of the message entirely since Blockbench shows it by default when installing plugins.

SirJain0:
@JannisX11 Sorry, accidentally closed the PR but it should be good now. I got rid of the quick-message to allow BB to do it automatically.

PR #316, opened 2022-12-28
Add Backup Viewer plugin

ewanhowell5195:
This plugin adds a backup tab to the Blockbench start screen that allows you to view a list of the project backups from within Blockbench.


JannisX11:
* The addition in plugins.json should use tab indentation for consistency
* I would put the accent color line for the tabs under the text, that way it's consistent with the icon size buttons on the right.

That's all the feedback I have. Great plugin!

PR #318, opened 2023-01-07
Add BBS exporter plugin

mchorse:
This plugin adds an exporter option to export current model (in any project format) to my BBS engine.

### DIFF
--- a/plugins/bbs_exporter.js
+++ b/plugins/bbs_exporter.js
@@ -1,0 +1,368 @@
+(function() {
+    var electron = require("electron");
+    var button;
+    var lastOptions = {};
+    var sides = {
+        north: "front",
+        south: "back",
+        west: "left",
+        east: "right",
+        up: "top",
+        down: "bottom"
+    };
+
+    /* Model exporter code */
+
+    function createHierarchy(objects)
+    {
+        var groups = {};
+
+        for (let i = 0; i < objects.length; i++)
+        {
+            var object = objects[i];
+
+            if (object.type === "group")
+            {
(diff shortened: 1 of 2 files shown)


plugins/bbs_exporter.js
```diff
@@ -0,0 +1,368 @@
...
+                if (formData.copyOnlyFirst)
+                {
+                    data = compileFirstCubes();
+                }
+                else
+                {
+                    data = compile();
+                }
+
+                electron.clipboard.writeText(autoStringify(data));
```
JannisX11, line 335: Use `Clipbench.setText(text)` to write to the clipboard, instead of requiring Electron. That way it will also work in the web app.

plugins/bbs_exporter.js
```diff
@@ -0,0 +1,368 @@
...
+
+        for (let i = 0; i < o.children.length; i++)
+        {
+            var object = o.children[i];
+
+            if (object.type === "group")
+            {
+                groups[object.name] = createModelGroup(object, groups);
+            }
+            else
```
JannisX11, line 58: You are forgetting to check if the element is a cube. It currently throws errors when the model contains locators etc.

JannisX11 (changes requested):

plugins/bbs_exporter.js
mchorse: Oh, I didn't know that, thanks!

plugins/bbs_exporter.js
mchorse: I added `.type == "cube"` check, thank you! Is it good now?

PR #320, opened 2023-01-09
Add Highlight mod shape generator

ThatGravyBoat:
Overview:
---
This plugin supports exporting to the Highlight(Resourceful Lib) mod shape format, allowing for non axis-aligned visual bounding boxes for blocks.

The mod can be found here: https://links.resourcefulbees.com/highlight

---

I don't know what your policy is for mod support as you have a plugin for Geckolib, Highlight the mod itself (which is used to provide vanilla compat) is a bit small but the JSON API behind it which is in [Resourceful Lib](https://www.curseforge.com/minecraft/mc-mods/resourceful-lib) is a bit bigger.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -723,5 +723,14 @@
 		"version": "1.0.0",
 		"variant": "both",
 		"tags": ["Exporter"]
+	},
+	"highlight_generator": {
+		"title": "Highlight Mod Shape Generator",
+		"icon": "fa-cube",
+		"author": "ThatGravyBoat",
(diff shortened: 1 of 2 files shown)


JannisX11:
Looks good. The only issue I noticed is that it doesn't generate a closing square bracket at the end of the file, resulting in an invalid JSON file.

ThatGravyBoat:
Sorry about that I must have by accident removed it when I put it in the plugins folder and cleaned up the code because the tab was still there, should be fixed now.

JannisX11:
Yup looks good.
I just found another issue, the three functions at the end are not wrapped into the main function. That can cause issues when side-loading the plugin.

PR #329, opened 2023-01-21
Add MC Text Generator plugin and edit Ground Plane Editor

SirJain0:
## Minecraft Text Generator

This plugin generates blocky, Minecraft-styled text out of cubes. Users can use this for logos, animations, and more.
To use, go to `Tools -> Generate Text` to generate text (making sure to leave no fields blank). The plugin works for all formats, but in the case of text exceeding the model bounds in certain formats, a warning will be shown.

## Ground Plane Editor

**Changelog**:
- Adjusted dialog width
- Changed name of action
- Added a button to reset dialog values
- Got rid of some unnecessary log statements 
- Action now does not show up in image mode

This PR also updates the code of my 5 other plugins, but with no functionality changes: I simply went and refactored a bit of the code, mainly making indenting consistent and replacing old code that had `var`, etc. And finally, it includes bugfixes for 4.6.3's dialog styling issues.


Malik12tree:
Really cool plugin! Some feedback I want to give,
1. There should be an option to create each letter in its own sub group, this would be really useful for animating!
1. Generated cubes should be all the same color for better UX (at least for cubes in each letter).
1. In the generate layer, "Your depth field needs to be 0 in order for the setting to work.", Perhaps you should add a condition to that field for it to only be available when depth equals zero.

Best regards!

SirJain0:
Hey @Malik12tree. Thanks for the suggestions! I'll try and implement these (if not now in a future version)

plugins/mc_text_generator.js
```diff
@@ -0,0 +1,711 @@
...
+        version: "1.0.0",
+        min_version: "4.2.0",
+        variant: "both",
+
+        oninstall: () => showAbout(true),
+        onload() {
+            addAboutButton()
+
+            // Define the dialog
+            const generateTextDialog = new Blockbench.Dialog({
```
JannisX11, line 36: Dialog ID is missing

plugins/mc_text_generator.js
```diff
@@ -0,0 +1,711 @@
...
+                name: "Generate Text",
+                icon: icon,
+                buttons: ["Generate", "Cancel"],
+                form: {
+                    input: {
+                        label: "Enter Text",
+                        type: "text",
```
JannisX11, line 43: My tip: Set `full_width: true` to get a bigger text input.
You could even add a custom CSS rule via Blockbench.addCSS that sets `text-transform` to `uppercase` on this input. That way users can directly see that all texts is converted to upper case letters.

plugins/mc_text_generator.js
```diff
@@ -0,0 +1,711 @@
...
+
+                        textGroup.children.forEach(group => {
+                            group.children.forEach(cube => {
+                                cube.moveVector(textLength / 2 - 4, 0)
+                            })
+                        })
+                        
+                        editBone()
+                        Canvas.updateView({groups: [textGroup, Group.selected], transform: true});
+                        Undo.finishEdit("Generated Text", {elements: selected});
```
JannisX11, line 562: The `outliner` and `group` aspects are missing in this undo statement, causing the redo function to not work correctly.

JannisX11 (commented):
I still need some more time to go over the other files. That's a lot of line changes...

SirJain0:
@JannisX11 Thanks for the thoughtful suggestions/feedback. I've implemented all of it. FYI, the line changes are a lot because for all my plugins files (besides Ground Plane Editor) I only fixed stuff like indenting, which changes a lot of lines.

plugins/mesh_flagger.js
```diff
@@ -1,178 +1,204 @@
...
+                    if (highlighter.running) return
+                    highlighter.running = true
+
+                    for (const cube of cubes) {
+                        cube.mesh.material_non_flash = cube.mesh.material
+                    };
+
+                    clearInterval(highlighter.interval)
+                    highlighter.i = 0
+                    highlighter.interval = setInterval(() => highlighter.flash(cubes, material), 1500)
```
JannisX11, line 46: It's not the best user experience to have to wait for 1.5 seconds for something to happen after clicking. You should start in the flashed state. And maybe make the inverval generally faster.
Same applies for the cube flash plugin.

SirJain0: I've fixed the empty flash at beginning part. My Utility Flaggers plugin which I will submit once this PR is merged will have customizable options for the duration, so I will leave that be for now.

plugins/ground_plane_editor.js
```diff
@@ -1,222 +1,251 @@
...
+                                }
+                            </style>
+                        `],
+                        component: {
+
+                            // Method runs when the dialog is opened
+                            mounted() {
+                                this.$nextTick().then(() => {
+                                    let opacityScaled = Math.floor(Canvas.ground_plane.material.opacity * 255);
+                                    $("dialog#edit_ground_plane_dialog .slider_input_combo #opacity_number").val(opacityScaled);
```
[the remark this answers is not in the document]
SirJain0: I'm not 100% sure on what this implements that I don't currently have in my plugin. Could you elaborate on that? (I probably missed something so my apologies)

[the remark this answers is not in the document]
JannisX11: You are currently using jQuery to set and read the values from the input. It would be much easier to use v-model.

You could also use a form by the way.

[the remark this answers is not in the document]
SirJain0: My current system does work fine. Unless you really feel like I should change, I don't think it's worth it for now.

PR #331, opened 2023-01-27
Add sam3DJ plugin

1Turtle:
This plugin allows you to export a project in to the .3DJ format.
It got introduced for the CC-Peripherals mod *(java edition)*, within the SwitchCraft3 Server. *(See more [here](https://docs.sc3.io/whats-new/sc-peripherals.html#_3dj-format))*

### DIFF
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -1,11 +1,6 @@
-//require('blockbench-types')
-
-// const { fstat } = require("fs");
-
 /** sam3DJ | aka. Sammy's .3DJ (format implementation)
  *  ==================================================
  *  Author:     Sammy L. Koch (1Turtle)
- *  GitHub:     1Turtle/sam3dj
  *  Licence:    Apache-2.0 (see https://www.apache.org/licenses/LICENSE-2.0.txt)
 */
 (function() {    
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -4,349 +4,340 @@
  *  Licence:    Apache-2.0 (see https://www.apache.org/licenses/LICENSE-2.0.txt)
 */
 (function() {    
-const path = require("path")
-
-const pluginProperties = {
-    title: 'sam3DJ',
-    author: '1Turtle',
-    tags: ['Minecraft: Java Edition', 'format', '3dj'],
-    version: '1.0.0',
-    variant: 'both',
-    description: 'Export models to the .3DJ format!',
-    about: `This plugin allows you to **export a model into the .3DJ format**.
-The .3DJ format is _mostly_ used with the \`3D Printer\` from the **CC-Peripherals** Java Edition mod.
-  \n
-> **Note:** _The format does not support the following atributes:_
-> * Rotated cubes
-> * One cube with multible textures
-> * Cubes outside the given block space *(16x16x16 Grid / 1x1x1 Block)*
-  \n
-Try to avoid those things, or else you might run into weird results.
-> For the full format documentations, see the [SwitchCraft3 Wiki](https://docs.sc3.io/whats-new/sc-peripherals.html#_3dj-format)`,
-    icon: 'print',
-    onload() {
-        MenuBar.addAction(btnExport, 'file.export');
-    },
-    onunload() {
-        MenuBar.removeAction('file.export.export_3dj');
-    }
-};
-const SC_TEXTURE_BLANK = 'sc-peripherals:block/white';
-
-/**
- * Returns a fitting texture for a given cube.
- * @param {*} cube The cube, which holds the wanted texture.
- * @returns The wanted texture, fitting to the cube. (+ some booleans) in a object.
- */
-function getSingleTexture(cube) {
-    let faces = 0;
-    let missing = false;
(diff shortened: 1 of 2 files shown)


Malik12tree:
Perhaps you should use the default form builder than using custom lines.
Check https://www.blockbench.net/wiki/api/dialog

1Turtle:
I dont quiet understand. I created the form's just like the wiki told.
I cant find an example for a form builder there too. How would I do that?

Malik12tree:
Mentioned in the wiki (modified):
```js
new Dialog({
    id: 'generate_tree',
    title: 'Tree Generator',
    form: {
        branches: {label: 'Branches', type: 'number', value: 10, step: 1, min: 1, max: 16},
        conifer: {label: 'Conifer', type: 'checkbox'},
    },
    onConfirm: function(formData) {
        console.log(formData.branches);
        console.log(formData.conifer);
    }
}).show()
```

1Turtle:
Oh I see now. You mean that I should only pass the Object to new Dialog, instead of the string too?

SirJain0:
image
Small typos here. (Make sure to update it in `plugins.json` too)

plugins/sam3dj.js
```diff
@@ -0,0 +1,343 @@
...
+*/
+(function() {    
+    const pluginProperties = {
+        title: 'sam3DJ',
+        author: '1Turtle',
+        tags: ['Minecraft: Java Edition', 'format', '3dj'],
+        version: '1.0.0',
+        variant: 'both',
+        description: 'Export models to the .3DJ format!',
+        about: `This plugin allows you to **export a model into the .3DJ format**.  \n
```
JannisX11, line 14: Please write this all in one line, like in plugins.json. The indentation confuses the markdown renderer, causing the text to be displayed as code.

plugins/sam3dj.js
```diff
@@ -0,0 +1,343 @@
...
+    /**
+     * Generates the object that contains the shapes in .3DJ format (off state only).
+     * @param {*} objects Contains all elements.
+     * @returns The final shapes (+ some booleans) in a object.
+     */
+    function genModel(elements, raw = false) {
+        let shapeOFF = [];
+        let shapeON = [];
+    
+        for (const cube of elements)
```
JannisX11, line 114: This assumes that all elements are cubes, and if that's not the case, it will run into an unhandled exception. You should check for the type here.

plugins/sam3dj.js
```diff
@@ -0,0 +1,343 @@
...
+     * Checks if all given cubes meet the requirements.
+     * @param {Cube} elements The cubes.
+     * @returns Object containing booleans for whenever a condition is met.
+     */
+    function checkCubes(elements) {
+        let rotation = false;
+        let texture = false;
+        let border = false;
+    
+        for (const cube of elements) {
```
JannisX11, line 256: This also assumes that all elements are cubes.

plugins/sam3dj.js
```diff
@@ -0,0 +1,343 @@
...
+ *  Licence:    Apache-2.0 (see https://www.apache.org/licenses/LICENSE-2.0.txt)
+*/
+(function() {    
+    const pluginProperties = {
+        title: 'sam3DJ',
+        author: '1Turtle',
+        tags: ['Minecraft: Java Edition', 'format', '3dj'],
+        version: '1.0.0',
+        variant: 'both',
+        description: 'Export models to the .3DJ format!',
```
JannisX11, line 13: Please include in the description the purpose of the file format, so that users have more info than just the file extension to evaluate whether this is the right plugin for them.

plugins/sam3dj.js
```diff
@@ -0,0 +1,343 @@
+/** sam3DJ | aka. Sammy's .3DJ (format implementation)
+ *  ==================================================
+ *  Author:     Sammy L. Koch (1Turtle)
+ *  Licence:    Apache-2.0 (see https://www.apache.org/licenses/LICENSE-2.0.txt)
+*/
+(function() {    
+    const pluginProperties = {
+        title: 'sam3DJ',
+        author: '1Turtle',
+        tags: ['Minecraft: Java Edition', 'format', '3dj'],
```
JannisX11, line 10: Let's include "Exporter" in the tags. I haven't written guidelines for tags yet, but that is planned, and should help make the tags more consistent across plugins.

JannisX11 (changes requested):
> Oh I see now. You mean that I should only pass the Object to new Dialog, instead of the string too? Yeah I can do that.

That's okay. Both ways to create dialogs are supported. I'm not sure what Malik meant by "lines" to be honest.

I do have some notes though:

1Turtle:
I see, I will implement the requested changes.
Thanks for pointing them out!

1Turtle:
> > Oh I see now. You mean that I should only pass the Object to new Dialog, instead of the string too? Yeah I can do that.
> 
> That's okay. Both ways to create dialogs are supported. I'm not sure what Malik meant by "lines" to be honest.
> 
> I do have some notes though:

I have implemented the above written points with this commit: fef792c9917a70b92f5981489da4105c1d61489d

JannisX11 (approved):

PR #338, opened 2023-02-20
Fix for on states

RPMYT:
Without this fix, the plugin does not work (at all).


RPMYT:
edit: this only sort of works. making a new PR.

PR #339, opened 2023-02-21
Some QoL changes and fixes to Sam3DJ

RPMYT:
This PR adds a new "feature" (if you can call it that) and fixes a bug with on states that prevented exports.
The new "feature" is that if you name a cube a block texture (i.e. if the name contains `block:/`) it will use that texture on export. This also works for groups, so you no longer need to manually change textures to Minecraft ones. (It'd be nice if it actually rendered as the texture, but I don't know _nearly_ enough about JS and Blockbench's API to do that)
The on-state bug fix is simply to not recursively check on states, which _does_ unfortunately have the side effect of not allowing sub-groups to be on states; however personally I'd rather that than have the plugin not work at all.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3,7 +3,7 @@
 		"title": "sam3DJ",
 		"author": "1Turtle",
 		"tags": ["Minecraft: Java Edition", "Format", "Exporter"],
-		"version": "1.0.1",
+		"version": "1.0.2",
 		"variant": "both",
 		"description": "Generate 3D prints in MC (Java Edition) via the .3DJ format within the SC-Peripherals mod!",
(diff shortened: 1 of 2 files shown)


JannisX11:
The plugin version in the plugin file and in plugins.json would need to be increased.
@1Turtle what do you think?

1Turtle:
Yeah it would be good to update the version then too, as those are minor changes.

1Turtle:
I think I see now why the `on_state` bug happend. *(I btw. didn't know there was one in the first place, a Issue would've been helpful.)*

I forgot to return false there, which is why it give nothing sometimes then probably. It should be a pretty simple fix to put a `return false` there at the end, if both if cases don't happen.

I would still prefer it being reqursive, but how you changed it for now is not bad too.

RPMYT:
Updated version in plugins.json and in plugin file (totally didn't forget about the plugin file)

RPMYT:
@JannisX11 @1Turtle is this ready to be merged?

1Turtle:
Yes this seems fine for me

PR #342, opened 2023-02-27
fix Export offset, Make defining textures easier

Merith-TK:
This has a few features,
Namely you can name your group or cube to apply an texture (will edit this message with screenshots)

and of course fix the `-8` offset that Blockbench has, previously the plugin *incorrectly* assumed `x0,y0,z0` was the origin (center) of the workspace, while it is actually `-8x, 0y, -8z`. this adjusts the values when exporting to contain an +8 to properly center it

The reason for the lines being all altered is because my editor uses tab indentations and using "format document" made them all consistent as there was a mix of 4space and tab indents everywhere, 

here are the diff files between my patches (after the indentation was changed) For the Texturing
```diff
diff --git a/sam3dj.js b/sam3dj.js
index 0e09b71..0092094 100644
--- a/sam3dj.js
+++ b/sam3dj.js
@@ -34,7 +34,6 @@
 
 		for (let dir in cube.faces) { //in ['north', 'east', 'south', 'west', 'up', 'down']
 			let tmp = cube.faces[dir].getTexture();
-
 			if (!tmp) {
 				if (typeof cube.parent === 'object' && cube.parent.name.includes(":block/")) {
 					texture = cube.parent.name;
@@ -50,14 +49,48 @@
 					missing = true;
 				}
 			} else {
-				// Generate actual ID
-				let id = tmp.namespace + ':' + tmp.folder + '/' + (tmp.name.replace(/\.[^/.]+$/, ""))
+				let id = ""
+				// if cube is titled <group>:<path>, use that as texture
+                if (cube.name.includes(":")) {
+                    id = cube.name.replace(/\.[^/.]+$/, "")
+                } else {
+					// if cube is in a group, use root group as namepsace
+					// use the rest of the path as texture name
+                    if (typeof cube.parent === 'object') {
+						// get all parents
+						let parents = []
+						let parent = cube.parent
+						while (typeof parent.parent === 'object') {
+							parents.push(parent.name)
+							parent = parent.parent
+						}
+						parents.push(parent.name)
+						parents.reverse()
+						// use the first parent as group
+						let group = parents[0]
+						parents.shift()
+						// if the group is not a folder, use the cube name as texture name
+						if (!group.includes(":")) {
+							parents.push(cube.name.replace(/\.[^/.]+$/, ""))
+						}
 
-				if (texture !== id)
-					faces++;
-				texture = id;
-			}
-		}
+						// set id to group:path/to/cube
+						id = group + ":" + parents.join("/") //+ "/" + cube.name.replace(/\.[^/.]+$/, "")
+
+
+					} else {
+						// if cube is not in a group, use the cube name as texture name
+                        id = "minecraft:" + tmp.name.replace(/\.[^/.]+$/, "")
+						console.log(id+" has no parent")
+                    }
+                }
+
+                console.log("Exporting " + cube.name + " as " + id)
+                if (texture !== id)
+                    faces++;
+                texture = id;
+            }
+        }
 
 		return {
 			texture: texture,
```

For the Offset
```diff
diff --git a/sam3dj.js b/sam3dj.js
index 0092094..88305d6 100644
--- a/sam3dj.js
+++ b/sam3dj.js
@@ -51,12 +51,12 @@
 			} else {
 				let id = ""
 				// if cube is titled <group>:<path>, use that as texture
-                if (cube.name.includes(":")) {
-                    id = cube.name.replace(/\.[^/.]+$/, "")
-                } else {
+				if (cube.name.includes(":")) {
+					id = cube.name.replace(/\.[^/.]+$/, "")
+				} else {
 					// if cube is in a group, use root group as namepsace
 					// use the rest of the path as texture name
-                    if (typeof cube.parent === 'object') {
+					if (typeof cube.parent === 'object') {
 						// get all parents
 						let parents = []
 						let parent = cube.parent
@@ -80,17 +80,17 @@
 
 					} else {
 						// if cube is not in a group, use the cube name as texture name
-                        id = "minecraft:" + tmp.name.replace(/\.[^/.]+$/, "")
-						console.log(id+" has no parent")
-                    }
-                }
-
-                console.log("Exporting " + cube.name + " as " + id)
-                if (texture !== id)
-                    faces++;
-                texture = id;
-            }
-        }
+						id = "minecraft:" + tmp.name.replace(/\.[^/.]+$/, "")
+						console.log(id + " has no parent")
+					}
+				}
+
+				console.log("Exporting " + cube.name + " as " + id)
+				if (texture !== id)
+					faces++;
+				texture = id;
+			}
+		}
 
 		return {
 			texture: texture,
@@ -108,13 +108,12 @@
 		let result = getSingleTexture(cube);
 		let shape = {
 			bounds: [
-				cube.from[0],
+				cube.from[0] + 8,
 				cube.from[1],
-				cube.from[2],
-
-				cube.to[0],
+				cube.from[2] + 8,
+				cube.to[0] + 8,
 				cube.to[1],
-				cube.to[2],
+				cube.to[2] + 8,
 			],
 			texture: result.texture,
 		};
@@ -301,12 +300,23 @@
 						texture = (getSingleTexture(cube).multipleTexture);
 
 					if (!border) {
-						function checkNum(num) {
+						// NOTE: If you can find a better way to do this, please do so.
+						// These changes had to be done due to Blockbench's 0,0 origin
+						// being in the center of the block and not in the bottom left corner.
+						// causing an malformed export. (off set by 8 in the x and z axis)
+						function checkNum(num, height) {
 							if (!border)
-								border = (num < 0 || num > 16);
+								if (height)
+									border = (num < -16 || num > 16);
+								else
+									border = (num < -8 || num > 8);
 						}
-						cube.from.forEach(checkNum)
-						cube.to.forEach(checkNum)
+						checkNum(cube.from[0]);
+						checkNum(cube.from[1], true);
+						checkNum(cube.from[2]);
+						checkNum(cube.to[0]);
+						checkNum(cube.to[1], true);
+						checkNum(cube.to[2]);
 					}
 
 					// We are done here
```

### DIFF
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -62,19 +62,22 @@
 						parents.reverse()
 						namespace = parents[0]
 						parents.shift()
-
-						logOutput = "Exporting " + cube.name + " in " + namespace + "/" + parents.join("/") + " as " + namespace + ":" + parents.join("/") + "/" + cube.name
-						id = namespace + ":" + parents.join("/") + "/" + cube.name
+						path = parents.join("/")
+						if (path !== "") {
+							path += "/"
+						}
+						id = namespace + ":" + path + cube.name
+						logOutput = "Exporting " + cube.name + " in " + namespace + "/" + path + " as " + id
 					} else {
 						if (cube.name.includes(":")) {
-							logOutput = "Exporting " + cube.name + " as " + cube.name
 							id = cube.name
+							logOutput = "Exporting " + cube.name + " as " + id
 						} else if (cube.name.includes("/")) {
-							logOutput = "Exporting " + cube.name + " as minecraft:" + cube.name
 							id = "minecraft:" + cube.name
+							logOutput = "Exporting " + cube.name + " as " + id
 						} else {
-							logOutput = "Exporting " + cube.name + " as minecraft:block/" + cube.name
 							id = "minecraft:block/" + cube.name
+							logOutput = "Exporting " + cube.name + " as " + id
 						}
 					}
 				}
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -27,70 +27,74 @@
 	 * @param {*} cube The cube, which holds the wanted texture.
 	 * @returns The wanted texture, fitting to the cube. (+ some booleans) in a object.
 	 */
-	function getSingleTexture(cube) {
+	function getSingleTexture(cube, scan) {
 		let faces = 0;
 		let missing = false;
 		let texture = SC_TEXTURE_BLANK;
 
 		for (let dir in cube.faces) { //in ['north', 'east', 'south', 'west', 'up', 'down']
 			let tmp = cube.faces[dir].getTexture();
-			if (!tmp) {
-				if (typeof cube.parent === 'object' && cube.parent.name.includes(":block/")) {
-					texture = cube.parent.name;
-					tmp = true;
-				}
-
-				if (cube.name.includes(":block/")) {
-					texture = cube.name;
-					tmp = true;
-				}
-
-				if (!tmp) {
-					missing = true;
-				}
-			} else {
-				let id = ""
-				// if cube is titled <group>:<path>, use that as texture
+			if (tmp === undefined) {
+				missing = true;
+				continue;
+			}
+			let logOutput = ""
+			// texture: <namespace>:<path>
+			let namespace = ""
+			let id = ""
+			if (tmp.name.includes(":")) {
+				id = tmp.name
+			} else if (tmp.name.includes("/")) {
+				id = "minecraft:" + tmp.name
+			}
+			if (id === "") {
 				if (cube.name.includes(":")) {
 					id = cube.name.replace(/\.[^/.]+$/, "")
+					logOutput = "Exporting " + cube.name + " as " + id
 				} else {
-					// if cube is in a group, use root group as namepsace
-					// use the rest of the path as texture name
 					if (typeof cube.parent === 'object') {
-						// get all parents
 						let parents = []
 						let parent = cube.parent
-						while (typeof parent.parent === 'object') {
+						while (typeof parent === 'object') {
 							parents.push(parent.name)
 							parent = parent.parent
 						}
-						parents.push(parent.name)
 						parents.reverse()
-						// use the first parent as group
-						let group = parents[0]
+						namespace = parents[0]
 						parents.shift()
-						// if the group is not a folder, use the cube name as texture name
-						if (!group.includes(":")) {
-							parents.push(cube.name.replace(/\.[^/.]+$/, ""))
+
+						if (scan) {
+						logOutput = "Exporting " + cube.name + " in " + namespace + "/" + parents.join("/") + " as " + namespace + ":" + cube.name
 						}
-
-						// set id to group:path/to/cube
-						id = group + ":" + parents.join("/") //+ "/" + cube.name.replace(/\.[^/.]+$/, "")
-
-
+						id = namespace + ":" + cube.name
 					} else {
-						// if cube is not in a group, use the cube name as texture name
-						id = "minecraft:" + tmp.name.replace(/\.[^/.]+$/, "")
-						console.log(id + " has no parent")
+						if (cube.name.includes(":")) {
+							
+							logOutput = "Exporting " + cube.name + " as " + cube.name
+							id = cube.name
+						} else if (cube.name.includes("/")) {
+							logOutput = "Exporting " + cube.name + " as minecraft:" + cube.name
+							id = "minecraft:" + cube.name
+						} else {
+							logOutput = "Exporting " + cube.name + " as minecraft:block/" + cube.name
+							id = "minecraft:block/" + cube.name
+						}
 					}
 				}
+			} else {
+				logOutput = "Exporting " + cube.name + " as " + id
+			}
 
-				console.log("Exporting " + cube.name + " as " + id)
-				if (texture !== id)
-					faces++;
+			if (scan && logOutput !== "") {
+				console.log(logOutput)
+			}
+		
+			if (texture !== id) {
+				faces++;
 				texture = id;
 			}
 		}
+
 
 		return {
 			texture: texture,
@@ -98,6 +102,7 @@
 			multipleTexture: (faces > 1),
 		};
 	}
+
 
 	/**
 	 * Converts a given element to a shape in .3DJ format.
@@ -297,7 +302,7 @@
 							|| cube.rotation[2] !== 0);
 
 					if (!texture)
-						texture = (getSingleTexture(cube).multipleTexture);
+						texture = (getSingleTexture(cube, true).multipleTexture);
 
 					if (!border) {
 						// NOTE: If you can find a better way to do this, please do so.
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -63,13 +63,10 @@
 						namespace = parents[0]
 						parents.shift()
 
-						if (scan) {
-						logOutput = "Exporting " + cube.name + " in " + namespace + "/" + parents.join("/") + " as " + namespace + ":" + cube.name
-						}
-						id = namespace + ":" + cube.name
+						logOutput = "Exporting " + cube.name + " in " + namespace + "/" + parents.join("/") + " as " + namespace + ":" + parents.join("/") + "/" + cube.name
+						id = namespace + ":" + parents.join("/") + "/" + cube.name
 					} else {
 						if (cube.name.includes(":")) {
-							
 							logOutput = "Exporting " + cube.name + " as " + cube.name
 							id = cube.name
 						} else if (cube.name.includes("/")) {
@@ -88,7 +85,7 @@
 			if (scan && logOutput !== "") {
 				console.log(logOutput)
 			}
-		
+
 			if (texture !== id) {
 				faces++;
 				texture = id;
--- a/plugins/sam3dj.js
+++ b/plugins/sam3dj.js
@@ -3,351 +3,394 @@
  *  Author:     Sammy L. Koch (1Turtle)
  *  Licence:    Apache-2.0 (see https://www.apache.org/licenses/LICENSE-2.0.txt)
 */
-(function() {    
-    const pluginProperties = {
-        title: 'sam3DJ',
-        author: '1Turtle',
-        tags: ['Minecraft: Java Edition', 'Format', 'Exporter'],
-        version: '1.0.2',
-        variant: 'both',
-        description: 'Generate 3D prints in MC (Java Edition) via the .3DJ format within the SC-Peripherals mod!',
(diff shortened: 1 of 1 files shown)


Merith-TK:
Before offset Patch
image
After offset patch 
image


example of how the texture labeling system works CURRENTLY (untested with on/off states as I am unaware as to how that works)

Three examples of how you can assign a texture (note, putting an `:` in your cube name overrides parent groups)
image
```json
{
	"label": "placeholder",
	"tooltip": "Made with Blockbench",
	"isButton": false,
	"collideWhenOn": true,
	"collideWhenOff": true,
	"lightLevel": 0,
	"redstoneLevel": 0,
	"shapesOff": [
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:block/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:static/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:bone/cube",
			"tint": "FFFFFF"
		}
	],
	"shapesOn": []
}
```

JannisX11:
@1Turtle What do you think?
I have no notes or concerns.
Of course the version number needs to be increased in the plugin file and in plugins.json.

JannisX11:
@1Turtle bump

1Turtle:
Hm Im not sure.
It is a great addition that you can change textures this way. But making it work only so does not feel right to me. I would make it optional _(you can also make it enabled by default if you really want)_.
Also, for such a feature, it might be good to mention this in the plugins description too, or else people will get confused why their textures suddenly dont work anymore.
And the version number chould be changed too of course, like JannisX11 mentioned.

About the fix: Good to know that the center is not 0,0,0, though its weird that the docs dont mention that.

Merith-TK:
> Hmm, I'm not sure. It's a great addition that you can change textures this way...

In its currently released state, defining textures does not work at all, it outputs `:/<texturename>`,
meaning if i name the texture `minecraft:stone`, it outputs `:/minecraft:stone`, which is an invalid texture within the 3DJ format

currently I do have it setup so if an cube is not within the predefined structure, it utilizes `minecraft:<texture>`, 
image
image

```json
{
	"bounds": [0, 0, 0, 16, 16, 16],
	"texture": "minecraft:placeholder",
	"tint": "FFFFFF"
}
```

you will note I do not automatically prefix textures, 
as part of the 3DJ format, it requires an full texture path, so `minecraft:stone` would not work, it would need `minecraft:block/stone`

hence why I went more or less with the format I did.

If you have any suggestions or tweaks I am open to hearing them

1Turtle:
> In its currently released state, defining textures does not work at all, it outputs `:/<texturename>`

I see. The problem you are encountering is that your texture does not have a namespace _(e.g. `minecraft` in your case)_.
It's not a user-friendly use case, if this is not obvious, or if it does not fix by itself, or whatever.

This for sure should get fixed. So for now, ignore the part where I said you should make it optional.
I have to encounter this another time.
But please implement the other points in my previous message. _(Documenting it, etc.)_

You also mentioned that you didn't test this with the on/off states, because you don't know how they work.
Basically: everything in the `state_on` folder will only be displayed when the model's state is on.
However, I think it currently is not compatible with recursive folders? Not sure though.

Other than that, I have nothing against this getting merged, thank you. ^^

Merith-TK:
> I see. The problem you are encountering is that your texture does not have a namespace (e.g. minecraft in your case).
It's not a user-friendly use case, if this is not obvious, or if it does not fix by itself, or whatever.

with the official plugin, if texture is `minecraft:stone`
```json
{
	"bounds": [-8, 0, -8, 8, 16, 8],
	"texture": ":/minecraft:stone",
	"tint": "FFFFFF"
}
```

I noticed the code does check for `:block/` but that feels wrong to me because what if someone wants to use a texture that does not exist inside the `block` catagory?

I will update my code to fix texture name output however, maybe even write a program to import textures from an "resourcepack" into a workspace

Merith-TK:
<https://mega.nz/file/etYzjZhR#lb2LVJNiI4ZsWvXyyg5w1HmWQoiNN_PuM9KGZX8OSig> bbmodel for this file, actual texture file not really important, just pick a placeholder texture if you dont want to use what I provide here

placeholder

image
image
image

```json
{
	"label": "testBlock",
	"tooltip": "Made with Blockbench",
	"isButton": false,
	"collideWhenOn": true,
	"collideWhenOff": true,
	"lightLevel": 0,
	"redstoneLevel": 0,
	"shapesOff": [
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:static/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:bone/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "minecraft:block/stone",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "minecraft:block/texture",
			"tint": "FFFFFF"
		}
	],
	"shapesOn": []
}
```

Merith-TK:
fixed typo
`modid/block/'cube'` now exports as 
`modid:block/'cube'` and not `modid:'cube'`
```{
	"label": "testBlock",
	"tooltip": "Made with Blockbench",
	"isButton": false,
	"collideWhenOn": true,
	"collideWhenOff": true,
	"lightLevel": 0,
	"redstoneLevel": 0,
	"shapesOff": [
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:block/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:static/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "modid:/bone/cube",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "minecraft:block/stone",
			"tint": "FFFFFF"
		},
		{
			"bounds": [0, 0, 0, 16, 16, 16],
			"texture": "minecraft:block/texture",
			"tint": "FFFFFF"
		}
	],
	"shapesOn": []
}
```

JannisX11:
The plugin version number still isn't increased, as mentioned above.
I also noticed that it fails and throws an error when exporting cubes without a texture.
Also 1Turtle mentioned changed to the plugin description that are not implemented yet.

Merith-TK:
I can easily fix the version number
It also never did throw the error of missing textures to begin with

And I honestly dont know how to explain how this works in an short and concise way to put in the description

1Turtle:
Alright I can do the description, it's fine ^^

I was testing things again now, and I really can't reproduce the bug with the textures. Also, it doesn't have an offset? Exporting it to .3dj and then 3d printing it works just fine for me. So taking the changes for the -8 offset is not really an option right now I think.

Maybe some kind help description or even video maybe is something this plugin needs? Or a more intuitive way, that would be even better. This plugin also still needs tint support... >.>

JTK222:
The offset does make sense sometimes.
Blockbench does only apply the offset to entity based model formats.
Java Block & Item format, does not have that offset.
So the plugin should either lock you to a specific format, or account for the formats offset.
Considering of the no rotation limit, it might also be worth to consider making a custom format for this within Blockbench.

And regarding the description, Blockbench does have a help menu, it is always a possibility to just add an entry, and display a more complex help page, or link to an external one.

1Turtle:
Oh alright, thank you for clarifying! Yeah I guess making a proper format for it would be reasonable. I don't know if I can manage this however, so someone would need to make a PR.

plugins/sam3dj.js
```diff
@@ -1,353 +1,401 @@
...
+					}
+				if (result.texture)
+					warning.form.texture = {
+						type: 'info',
+						text: '... use multiple textures',
+					}
+				if (result.border)
+					warning.form.border = {
+						type: 'info',
+						text: '... are out of bounce (16x16x16 Grid / 1x1x1 Block)',
```
JTK222, line 380: typo,  "bounce"

plugins/sam3dj.js
```diff
@@ -1,353 +1,401 @@
...
+
+	/**
+	 * Converts a given element to a shape in .3DJ format.
+	 * @param {*} obj The given element (e.g. Cube).
+	 * @returns The fianl shape (+ some booleans) in a object.
+	 */
+	function genShape(cube, raw = false) {
+		let result = getSingleTexture(cube);
+		let shape = {
+			bounds: [
```
JTK222, line 115: Like I said in my comment, the offset depends on the model format, this should be accounted for, as otherwise users may run into issues.

plugins/sam3dj.js
```diff
@@ -1,353 +1,401 @@
...
+			// texture: <namespace>:<path>
+			let namespace = ""
+			let id = ""
+			if (tmp.name.includes(":")) {
+				id = tmp.name
+			} else if (tmp.name.includes("/")) {
+				id = "minecraft:" + tmp.name
+			}
+			if (id === "") {
+				if (cube.name.includes(":")) {
```
JTK222, line 51: This whole section seems questionable to me, if this condition is met, the other renaming logic further down does never run.
I assume file extension removal should always be performed, and be further down, with the other renaming logic.
Though I am not farmiliar with the 3DJ format, therefore please correct me if I am incorrect in my assumptions.

Merith-TK:
> I was testing things again now, and I really can't reproduce the bug with the textures. Also, it doesn't have an offset? Exporting it to .3dj and then 3d printing it works just fine for me. So taking the changes for the -8 offset is not really an option right now I think.

the Export Offset is because the printer mod (unless this behavior has been fixed in newer updates) uses 0,0 as the origin (corner) point, while in BlockBench 0,0 is used as the center point. the offset fixes that by moving the object off of blockbench's origin and to the 3dj format's origin


> This whole section seems questionable to me, if this condition is met, the other renaming logic further down does never run.
I assume file extension removal should always be performed, and be further down, with the other renaming logic.
Though I am not farmiliar with the 3DJ format, therefore please correct me if I am incorrect in my assumptions.

there are a few ways in which you can name the textures, as shown in this screenshot
image


The idea is that the user can define an *valid* minecraft texture path through naming the cube or by naming the texture they have applied to the cube, (texture name takes priority on an cube-by-cube basis)
and by default, if an texture is unable to be gathered by the program, default to stone so that the model does not have an missing texture in game

JTK222:
> > I was testing things again now, and I really can't reproduce the bug with the textures. Also, it doesn't have an offset? Exporting it to .3dj and then 3d printing it works just fine for me. So taking the changes for the -8 offset is not really an option right now I think.
I'll gladly say it a 3rd time, whether Blockbench has 0, 0, at the center, or in the corner. depends solely on the model format you're using. When you're modelling in block/item mode, this will create the exact issue you're trying to solve.

Merith-TK:
ah, missunderstood what you were saying there. could add in a check for the model type and apply the offset when in broken modes

for texture namings, `block.png` wouldn't be valid as it contains no formatting, so it will default as `minecraft:block/stone` for the texture

as for an cube or texture named as `:`, yeah but thats an edge case that is extremely unlikely to happen as no sane person to my knowledge (even the redstone maniacs) would name something when modeling "`:`", an check can be coded in for regex matching however if that will ease your concern?

also, texture name will take priority over cube name, 

in the attached image,
cube has `block.png` attached
`:` has no texture --

aigth in testing I have discovered that having an cube named `:` with no texture assigned, causes an error
image

PR #348, opened 2023-03-14
Added BAMO

tmudway:


JannisX11:
I reviewed your submission. Two notes:
* Blockbench already comes with JSZip by default, so you should treat it as an external dependency.
* Please include the files that are necessary to reproduce the build, such as webpack config and package.json, in the source folder.

tmudway:
> 

Both changes made. Anything else that looks off?

SirJain0:
@tmudway Any news on how this is going? Just curious since this seems like a cool plugin

tmudway:
Updated with requested changes

JannisX11:
Sorry this has been in the queue for so long!

JannisX11:
Not blocking, but I would recommend including the link to documentation in a clickable way. To do that, you can use the "website" metadata field and you can additionally provide an about text.

PR #351, opened 2023-03-18
Add Axis Color Customizer plugin

SirJain0:
## Axis Color Customizer
Adds a method for customizing the x, y, and z axis colors for **Blockbench purposes only**. Updates the Display tab slider colors, icons where applicable, and canvas gizmos. Use in `Tools > Customize Axis Colors`.

Why is this plugin useful? Some people may be used to flipped z and y axis colors, and this plugin can help those people get used to that. Also, (I haven't looked into this but some people have suggested it), it can be good for special vision for certain people to adjust the axis colors a bit.

I also fixed a bit of formatting in my MC Text Generator plugin, but that's nothing major.


SirJain0:
Above commits are just for reverting some stuff I did with MC Text Generator. Ignore that.

JannisX11:
Customizing UI colors is a default feature of Blockbench. Should be better documented maybe. But I don't think there needs a separate plugin just for the axis colors.
image

SirJain0:
Yeah better documentation would be appreciated. Thanks anyways.

PR #355, opened 2023-03-30
Activity Tracker plugin

ewanhowell5195:
This plugin allows you to keep track of how much time you spend using Blockbench and how long you spend working on each project.


JannisX11:
If I install the plugin while having a project open, and then open the dialog, it'll display NaN in the project section.
That's a caveat with properties, initializing one does not automatically apply their default value to all instances.

PR #357, opened 2023-04-01
Add "Animation to JsonEA Converter"

Gaming32:
This is a converter from Blockbench animations to the [JsonEA](https://modrinth.com/mod/json-entity-animation) format.


JannisX11:
Looks like I forgot to note this about the plugin that this was based on. But please make sure the functions at the bottom are inside the self-invoking function (before the `})();`) so that they don't bloat the global object.

Gaming32:
Done

PR #358, opened 2023-04-03
BBS model exporter: Fix keyframe exporter mishandling Effect keyframes

mchorse:
I didn't considered Effect keyframes, so this caused an error that doesn't allows exporting a model. I also added appending `.bbs` to the file name, because BBS expects its models to be have that extension (to avoid future configuration files that are just `.json`).

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -710,8 +710,8 @@
 		"title": "BBS model exporter",
 		"icon": "fa-file-export",
 		"author": "McHorse",
-		"description": "Adds a model exporter which allows to export models in BBS model format",
-		"version": "1.1.1",
+		"description": "Export model as a BBS (.bbs.json) model",
+		"version": "1.1.2",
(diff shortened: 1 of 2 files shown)


JannisX11:
There is now a description mismatch between plugins.json and the plugin file.
In general it might be good if the description explains what the models are for, since BBS is a fairly generic and not really googlable acronym.

mchorse:
> There is now a description mismatch between plugins.json and the plugin file. In general it might be good if the description explains what the models are for, since BBS is a fairly generic and not really googlable acronym.

Thank you for feedback! I updated and synced the description.

JannisX11:
Oops, you updated the description of the tool, not that of the plugin.

mchorse:
> Oops, you updated the description of the tool, not that of the plugin.

Oh... my bad, I fixed it! 😅

PR #363, opened 2023-04-18
MTools and Plugin Statistics Update

Malik12tree:
### New Features
- Add Per Week Downloads toggle
- Expand to yearly downloads
- Now the graph shows data starting from the plugin release week
### Fixes
- UV Project from View now uses vertices' world coordinates. #360 
- Expand/Shrink Selection now use new selection system.
- Fix some XYZ Math Surface Function functions


JannisX11:
minVersion should be min_version

Malik12tree:
Should be good now!

PR #366, opened 2023-04-25
Fix azurelib plugin entry

AzureDoom:
Noticed I had an issue with my plugins.json entry (that's what I get for doing it via mobile!)


JannisX11:
There's a weird merge error, I'll just fix it on my end.

AzureDoom:
Thanks!

PR #369, opened 2023-05-04
MTools: Add Preserve Aspect option for Project From View

Malik12tree:


JannisX11:
@Malik12tree It looks like you accidentally changed `min_version` to `minVersion` in the plugin file. I'll ignore this for now though since that field is not very relevant within the plugin file.

PR #371, opened 2023-05-19
Update Animated Java to 0.3.9a

SnaveSutit:


JannisX11:
I saw that you already switched to the production build again, that's good for the download size, it would have been to large otherwise.
Please use numeric semver versions in the future. It doesn't matter much in this case specifically since it will still detect a version jump between .8 and .9, but generally this is not supported.

SnaveSutit:
Yeah I only realized I uploaded a dev build after I uploaded the 0.3.9a build. 😅
I'll reform to number only versions next fix

PR #372, opened 2023-05-20
Added sphere plugin

hamburger2506:
Added a sphere creator for minecraft java edition which allows for the creation of pixelated spheres for easy pixel art

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -799,5 +799,14 @@
 		"min_version": "4.0.0",
 		"await_loading": true,
 		"variant": "both"
+	},
+	"sphere": {
+		"title": "PixelArt Sphere Generator",
+		"author": "hamburger2506",
+		"icon": "circle",
+		"description": "A plugin to create pixelated spheres, since minecraft does not allow real spheres due to something to do with rotation. Also I needed a sphere that I could do pixel art on to make it look like a planet. (Feel free to update or republish this plugin if you\'d like since I probably won\'t do mayor updates to it anymore.)",
+		"version": "1.0.0",
+		"variant": "both",
+		"tags": ['Minecraft: Java Edition']
 	}
 }
--- a/plugins.json
+++ b/plugins.json
@@ -807,6 +807,6 @@
 		"description": "A plugin to create pixelated spheres, since minecraft does not allow real spheres due to something to do with rotation. Also I needed a sphere that I could do pixel art on to make it look like a planet. (Feel free to update or republish this plugin if you\'d like since I probably won\'t do mayor updates to it anymore.)",
 		"version": "1.0.0",
 		"variant": "both",
-		"tags": ['Minecraft: Java Edition']
+		"tags": ["Minecraft: Java Edition"]
 	}
 }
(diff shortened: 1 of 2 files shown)


Malik12tree:
MY feedback on this plugin :3
1. The `tags` key in a plugin's meta data should be an array of strings rather than a string, thus, should be `tags: ['Minecraft: Java Edition']`.
2. You need to add your plugin's meta data in `plugins.json`, Possibly:
```json
,{
  "title": "PixelArt Sphere Generator",
  "author": "hamburger2506",
  "icon": "circle",
  "description": "A plugin to create pixelated spheres, since minecraft does not allow real spheres due to something to do with rotation. Also I needed a sphere that I could do pixel art on to make it look like a planet. (Feel free to update or republish this plugin if you'd like since I probably won't do mayor updates to it anymore.)",
  "version": "1.0.0",
  "variant": "both",
  "tags": ["Minecraft: Java Edition"]
}
```
3. Instead of having the center default to `7.5, 7.5, 7.5`, it could be set as `0, 0, 0` by default when `Project.format.centered_grid` is true.
4. I would like to suggest an alternative approach. Currently, you're using a process called "voxelization" to draw a sphere, which can be computationally expensive as the radius increases.
Perhaps, draw the sphere mathematically by extending the Midpoint Circle Algorithm into 3D space.
Although there is no original method, my suggestion would be to use trigonometric functions to determine the radius for each latitude circle.

I hope you find these suggestions helpful,
Best regards!

hamburger2506:
Ty for the advice, by 1 I didn't evern notice I used doubled quotes, for 2 I am very new to coding and have no exact idea of what metadata is and how to put it in a json but I will defi look into it. for 3, I set it up so the sphere is 1 block in size by default but sure I'll look into adding such a setting. and for 4, oh that sounds like it is going to be a heck load of work. I will defi look into it but idk if I want to make this an actual large optimizes plugin, I just made it to solve some small problem I had and thought it might be usefull for the community. I'll look into some things but I'm very busy with exams. Also if anyone knows the solution to these problems and wants to finish this plugin I also wouldn't mind lol.

plugins.json
```diff
@@ -799,5 +799,14 @@
...
 		"variant": "both"
+	},
+	"sphere": {
+		"title": "PixelArt Sphere Generator",
+		"author": "hamburger2506",
+		"icon": "circle",
+		"description": "A plugin to create pixelated spheres, since minecraft does not allow real spheres due to something to do with rotation. Also I needed a sphere that I could do pixel art on to make it look like a planet. (Feel free to update or republish this plugin if you\'d like since I probably won\'t do mayor updates to it anymore.)",
+		"version": "1.0.0",
+		"variant": "both",
+		"tags": ['Minecraft: Java Edition']
```
Malik12tree: In the JSON meta data, `tags` should be `"tags": ["Minecraft: Java Edition"]`

Malik12tree (changes requested):

plugins.json
hamburger2506: I fixed that in the newest pull request?

JannisX11:
Sorry, I can't accept this plugin into the plugin store in good conscience. Feel free to publish it on the Discord server instead.

PR #374, opened 2023-05-28
Add Cube Inverter, update MC Text Gen and Outline Creator

SirJain0:
# Cube Inverter
A simple plugin that inverts cubes with a click of the button.

There are many cases in which inverted cubes are useful (besides outlines), but making them is not too convenient as of now. Also, UVs will get messed up, which requires manual fixing. This plugin fixes that.

To use this plugin, press the button that appears in the size fields of the cubes (located next to the inflation box). The action is also keybinded for extra convenience.

# MC Text Generator
**Changelog:**
- Reword some incorrect or vague descriptions for some of the text settings
- Make layer setting only appear when the depth setting is 0
- Got rid of max limit for depth of characters
- Changed default thickness of characters
- Reworded the empty text field warning's description
- Added new characters: `$`, `/`
- Mark plugin as deprecated

# Outline Creator
**Changelog:**
- Added support for meshes

### DIFF
--- a/plugins/cube_inverter.js
+++ b/plugins/cube_inverter.js
@@ -1,0 +1,174 @@
+(async function() {
+        let aboutAction, cubeInverterAction
+        const id = "cube_inverter"
+        const name = "Cube Inverter"
+        const icon = "swap_horiz"
+        const author = "SirJain"
+
+        // Used for about dialog
+        const links = {
+                website: "https://sirjain0.github.io/",
+                twitter: "https://twitter.com/SirJain2",
+                discord: "https://discord.gg/wM4CKTbFVN"
+        }
+
+        // Registers plugin data
+        Plugin.register(id, {
+                title: name,
+                icon,
+                author,
+                description: "Adds a button that inverts selected cube sizes.",
+                about: "This plugin adds a button that inverts the size values of each selected cube.\n## How to use\nTo use this plugin, go to the size values of the cube (located in edit mode) and click the `Invert Cubes` button located next to the Inflation value. The button is also keybinded with a default of `Shift + I`. You can add or remove the tool using the `Customize Toolbar` option.\n\nPlease report any bugs or suggestions you may have.",
+                version: "1.0.0",
+                min_version: "4.2.0",
+                variant: "both",
+                oninstall: () => showAbout(true),
+                onload() {
+                        addAbout()
+
+                        cubeInverterAction = new Action("cube_inverter_action", {
+                                name: "Invert Cubes",
+                                icon: icon,
+                                description: "Invert the values of all axes of selected cubes.",
+                                keybind: new Keybind({key: 'i', shift: true}),
+                                condition: () => Cube.selected.length,
+                                click: () => invertCubes()
+                        })
+
+                        Toolbars.element_size.add(cubeInverterAction)
+                },
+                onunload() {
+                        aboutAction.delete()
+                        cubeInverterAction.delete()
+                        MenuBar.removeAction(`help.about_plugins.about_${id}`)
+                        Blockbench.showQuickMessage("Uninstalled Cube Inverter plugin")
+                }
+        })
+
+        function invertCubes() {
+                Undo.initEdit({elements: Cube.selected, outliner: true});
+
+                for (const cube of Cube.selected) {
+
+                        // Handles size dimensions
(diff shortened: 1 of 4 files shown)


Malik12tree:
Basically the same as the [Outline Creator](https://www.blockbench.net/plugins/outline_creator) plugin?

SirJain0:
No.
The outline creator plugin specifically adds another element and makes it slightly bigger which creates that effect.
Mine changes the dimensions of the already-present cube, which you currently have to do manually.

SirJain0:
Just updated this PR to include changes to the Outline Creator and MC Text Generator plugins.
@GitWither Just pinging you in advance - could you please review the changes made to your plugin? The update just adds support for meshes. Thanks.

GitWither:
I'm ok with the changes to my plugin

JannisX11:
Please start making individual PRs per plugin. Unless it's just the same change across different plugins. It helps me manage and review them faster.

SirJain0:
Will do for the future.

SirJain0:
(Typo in above commit's name - meant to say `MC Text Generator: Add "/" Character`)

plugins/mc_text_generator.js
```diff
@@ -24,8 +24,8 @@
         author,
         description: "Generates Minecraft-styled text in cubes.",
         about: "This plugin adds a button under the `Tools` menu that allows you to generate Minecraft-like text.\n## How to use\nTo use this plugin, go to `Tools > Generate Text`. Simply enter some text, configure your settings how you like, and press `Generate`!\n\nPlease report any bugs or suggestions you may have.",
-        tags: ["Minecraft", "Font", "Generator"],
-        version: "2.0.0",
+        tags: ["Deprecated", "Minecraft", "Font"],
```
JannisX11, line 27: Tags are not updated in plugins.json

plugins/cube_inverter.js
```diff
@@ -0,0 +1,174 @@
...
+                        [cube.from, cube.to] = [cube.to, cube.from]
+
+                        // UV handling
+                        cube.faces.north.uv = cube.faces.south.uv
+                        cube.faces.south.uv = cube.faces.north.uv
+                        cube.faces.north.texture = cube.faces.south.texture 
+                        cube.faces.south.texture = cube.faces.north.texture
+                        cube.faces.north.cullface = cube.faces.south.cullface 
+                        cube.faces.south.cullface = cube.faces.north.cullface
+                        cube.faces.north.rotation += 180
```
JannisX11, line 63: You might want to make sure that the rotation stays within the permitted values 0, 90, 180, 270.
Modulo solution: `rotation = (rotation + 180) % 360`

plugins/outline_creator.js
```diff
@@ -101,7 +97,28 @@ function createOutline(outline_thickness) {
...
+            face.invert();
+        })
+
+        mesh.resize(outline_thickness * 2, 0, false, false, true);
+        mesh.resize(outline_thickness, 1, false, false, true);
+        mesh.resize(outline_thickness * 2, 2, false, false, true);
+        mesh.name = mesh.name + "_outline";
+    }
+
+    Canvas.updateAll();
```
JannisX11, line 121: Canvas.updateAll is quite inefficient and will be deprecated at some point. I'd recommend to use Canvas.updateView API instead, with all affected elements and the transform and geometry aspects: https://github.com/JannisX11/blockbench/blob/master/js/modeling/transform.js#L676

SirJain0:
@JannisX11 Reviewed and made changes, thanks

SirJain0:
Sorry, I realized I overcomplicated the process of the elements aspect - just fixed that

JannisX11:
The extra button in the element size toolbar concerns me. On default sidebar width it causes the toolbar to overflow, causing the entire layout below to shift. This can shift the outliner around when selecting things. That's not a great default user experience.
Try and add this button to another toolbar or menu instead.

SirJain0:
Good point. I've added the action to the tools tab now, which I think makes the most sense.
Also fixed the missing `about` information in the plugins.json.

SirJain0:
Realized I made an error with the modulo fix for cube UV rotations. Just fixed that.
@JannisX11 I think this should be good, but also let me know if you think another tab would be better for the action for the Cube Inverter plugin.

PR #386, opened 2023-07-01
Adds Texture Tint export option to sam3dj.js

Magicmaan:
Implements option in Dialog to circumvent normal behaviour, instead assigning the cubes top face texture colour as tint and defaulting the texture to "sc-peripherals:block/white"


JannisX11:
Looks good to me.
@1Turtle good to merge?

PR #389, opened 2023-07-07
Rename MC Text Generator plugin to Text Generator

SirJain0:


JannisX11:
No, changing plugin names is not an issue.

SirJain0:
Thanks

PR #401, opened 2023-08-02
Updated Figura Model Format plugin to 0.1.1

KitCat962:
Minor update changing the Project Type's name from "Figura" to "Figura Model" (community suggestion) and fully removed Particle and Sound keyframes instead of the rename patch job.

Additionally, the plugin has moved to the 4.8.0 plugin format

Currently I am assuming that "instructions no longer need to be stored in the plugin meta data" stated in the Blockbench Discord in the "4.8 API changes for plugin creators" post means that I do not need an `about` field in `plugins.json` for my plugin, as `about.md` will be used in it's stead for versions 4.8.0 and above. I am also assuming that plugins that have already updated to 4.8.0 only have the `about` field because of compatibility with 4.7.0. 
If these assumptions are incorrect at all, please let me know before merging and I'll fix it.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -837,7 +837,7 @@
 		"title": "Figura Model Format",
 		"author": "Katt (KitCat962)",
 		"icon": "icon.svg",
-		"description": "Create models for the Figura mod without unusable Blockbench features getting in the way.",
+		"description": "Create models for the Figura mod in a custom format that optimizes Blockbench to work with Figura models.",
 		"tags": ["Minecraft: Java Edition", "Figura"],
 		"version": "0.1.1",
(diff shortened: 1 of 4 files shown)


JannisX11:
That is correct, the about.md file replaces the about field in the plugin meta data. Plugins may still use that field for compatibility if they have a min_version below 4.8.0 and a plugin file directly in the plugins directory.
Please update the plugin description so that it does not potentially sound like something is wrong with Blockbench. For example:
"Create models for the Figura mod in a custom format that optimizes Blockbench to work with Figura models"

KitCat962:
Sorry about that. I've replaced it with your suggestion.

PR #406, opened 2023-08-08
Add Workspaces plugin

ewanhowell5195:
This plugin adds a workspaces system that allows you to store your recent projects in separate workspaces.


JannisX11:
I really don't like the plugin moving around all the thumbnail files in the background. Especially with a few hundred recent files in a list, it can take a few seconds to change. And I'm generally not a fan of letting plugins move files around in the app data directory.
If the issue was the automatic cleanup of the directory, I know it's hard to access, but I have another solution for that. It's still hacky, but I prefer it over this.
There's the `update_recent_project_thumbnail` event hook right before the cleanup. So just add all thumbnails that should exist to the `recent_projects` list, and then remove them again a milisecond after. That way it will know that they are still required.

PR #412, opened 2023-08-20
Add sphere generator plugin

doonv:
I added a create plugin which creates a sphere with the size & shape specified by the user.

I also added a basic "voxel optimizer" which combines certain cubes together to save on element count. It's not that good though.


SirJain0:
The plugin could definitely be useful to some people but my main concern is that new Blockbench users will use the plugin without knowing the implications of both the lag and the lack of optimization it could cause, both in Blockbench and in Minecraft if they decide to use it for that (especially for new modelers). Since Blockbench is kind of branded as a low-poly modeling program and not a voxel editor, I'm on the fence of whether it falls out of line of what the program should be used for (on the fence because I know this is a plugin and not a default feature). I do like the idea of warning the user before generating a sphere with a lot of elements though.

In any case, I have a couple of notes based on when I tested it:
- You seem to be missing the `Undo.initEdit` and `Undo.finishEdit` aspects in your plugin, meaning that users cannot undo the action if they make a sphere. You can read about those here: https://www.blockbench.net/wiki/docs/undo
- I'd maybe mention how the plugin generates spheres (via voxels) in the plugin description or about dropdown for clarity.

That being said those are only my suggestions - besides the missing undo aspects, the plugin works well from when I tested it.

doonv:
Oh ok. I'll implement the changes you suggested.

doonv:
Ok I renamed the plugin to "Voxel Sphere Generator", added Undo, and added the plugin to `plugins.json`.

ewanhowell5195:
This plugin is defining an unused function `onUninstall` in global scope

doonv:
oh yeah, lemme fix that

JannisX11:
Sorry, I can't accept this plugin into the plugin store in good conscience. Blockbench is not designed for voxels and this would be abused. Feel free to publish it on the Discord server instead.

PR #414, opened 2023-08-23
Fix text overflowing codeblock for azurelib_utils about.md

ewanhowell5195:
Fixes a small issue where in azurelib_utils the text was overflowing the codeblock in the about.md

Pinging @AzureDoom to let you know about this PR

PR #417, opened 2023-09-01
Azurelib Utils Rename/Update

AzureDoom:
- Changes the display name of the plugin from `Azure Utils` to `Azurelib Animator`
- Swaps the right and left sides of the armor template itself; this is just a name swap.
- Updates tooltips and button names to better dissociate AzureLib from the Geckolib plugin. 
- Keep Display Settings active in all three plugin types. 
     - This is part of a planned deprecation of the just Block/Item type to Entity/Block/Item, but the Block/Item is kept to prevent broken model files. This will be removed in the future whenever a breaking Blockbench update happens. 
- Adds a link to the wiki I am working on creating for the AzureLib mod.
- Cleans up the `about.md` as well, removing some unneeded indenting to make the fix from #414 look nicer (at least to me) since it doesn't overflow anymore. 
- Building now puts the built plugin into the proper plugins folder instead of the root plugins folder. 

If anything doesn't meet the proper standards or needs changed please let me know.


AzureDoom:
Just bumping this to see if anything is wrong with it?

PR #422, opened 2023-09-21
Add url fallback system & font variants

ewanhowell5195:
This adds a url fallback system to Minecraft Title Generator.

Previously, I moved it to jsdelivr, as that is more reliable than raw.githubusercontent, but as you know, that can be really slow to update.

Now it has an array of root URLs that it will try to load from, when it fails to load the first, it changes the root URL to the next in the array until it runs out, or sucessfully loads one.

also fix a single css issue

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -311,7 +311,7 @@
 		"author": "Ewan Howell",
 		"description": "Create Minecraft-styled title models!",
 		"tags": ["Minecraft", "Title", "Logo"],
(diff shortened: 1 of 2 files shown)


ewanhowell5195:
2nd commit adds support for font variants. this is a breaking change, so it is using a dev branch of the font sources.
once this update goes through, i will make another update taking it off the dev branch after i update the main branch.

PR #424, opened 2023-09-26
Update Minecraft Wiki link to new domain after fork

misode:
The Minecraft Wiki maintainers have decided to move away from Fandom. More information can be found here: https://minecraft.wiki/w/Minecraft_Wiki:Moving_from_Fandom.


JannisX11:
Feels a bit superfluous since it's just a code comment to reference the source of some data, realistically no one will ever click this link.

misode:
Fair enough, although I could argue that this link will be used to update these values after a future release, which would be unfortunate when the old wiki is no longer being updated.

PR #438, opened 2023-11-04
sam3dj: Fixed Exporting

1Turtle:
Exporting within cubes that didn't had textures caused to an uncaught error because of the last feature commit for tinted values instead of texture names. This has been fixed within this patch.

The version has been dumped to 1.0.3 as well.


khankul:
Updated the plugin manually, it's working! Thanks a lot

PR #442, opened 2023-11-12
Update Figura Model Format to 0.1.2

KitCat962:
Updates the plugin to 0.1.2 making the following changes:
* Changed the style of about.md
* fixed an error related to updateProjectUV
* Removed the ability to have groups with the same name as it broke duplicating an animation
* Removed "Render Order" as Figura ignores the property
* Added an action to get the scripting path to a ModelPart
* Added an action to bake Inverse Kinematics to raw rotation keyframes
* reorganized the plugin code to hopefully be more readable in the future

Is it normal to add changes directly to a fork of this repo? 
I've been making changes in a separate repo and only updating the fork when I increment the version/make a pull request.
However, I've noticed many other prs directly committing to a fork of this repo.

Should I continue making changes in a separate repo and only modifying the fork when pull requesting, or should I be making all of my changes in a fork of this repo and keeping all the commit history? Or does it not matter?

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -867,7 +867,7 @@
 		"icon": "icon.svg",
 		"description": "Create models for the Figura mod in a custom format that optimizes Blockbench to work with Figura models.",
 		"tags": ["Minecraft: Java Edition", "Figura"],
-		"version": "0.1.1",
+		"version": "0.1.2",
 		"min_version": "4.8.0",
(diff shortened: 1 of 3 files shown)


KitCat962:
Just had a single feature idea that I wanted to implement before merge. Ready to merge anytime

PR #456, opened 2023-12-10
Update animator.js

Command-Master:
Fixed the save_start icon and updated pack_format in the Java Item Model Animator plugin.


JannisX11:
Looks like you forgot to pull, there was already a version 1.0.1. Also, please add the version number to plugins.json, as that's now required for the plugin to update correctly.

JTK222:
Wouldn't it make a lot more sense to turn the pack version into a dropdown?
From what I can tell, pack format 18 is already outdated. (Latest is 22 for resource packs and 26 for datapacks)

PR #473, opened 2024-02-05
Update VoxelShape Generators to Actual MojMaps and added Yarn.

Nyfaria:
It was still using MCP class names, Mojang Class names are now Available for use (have been for a while), so I updated it to use the Mojang Class names, and Added Yarn.

### DIFF
--- a/plugins/voxel_shape_generator.js
+++ b/plugins/voxel_shape_generator.js
@@ -5,10 +5,10 @@
 
     Plugin.register('voxel_shape_generator', {
         title: 'VoxelShape Generators',
-        author: 'Spectre0987',
-        description: 'Generates Voxel Shapes (Supports Mojang Mappings and MCP)',
+        author: 'Spectre0987, Nyfaria',
+        description: 'Generates Voxel Shapes (Supports MojMaps, MCP and Yarn)',
         icon: 'bar_chart',
-        version: '0.1.0',
+        version: '0.2.0',
         variant: 'both',
         tags: ["Minecraft: Java Edition"],
         onload(){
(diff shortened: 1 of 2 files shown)


Nyfaria:
@Spectre0987 I made this PR and just want to let you know.

plugins/voxel_shape_generator.js
```diff
@@ -5,10 +5,10 @@
 
     Plugin.register('voxel_shape_generator', {
         title: 'VoxelShape Generators',
-        author: 'Spectre0987',
-        description: 'Generates Voxel Shapes (Supports Mojang Mappings and MCP)',
+        author: 'Spectre0987, Nyfaria',
+        description: 'Generates Voxel Shapes (Supports MojMaps, MCP and Yarn)',
         icon: 'bar_chart',
-        version: '0.1.0',
+        version: '0.2.0',
```
JannisX11, line 11: Make sure to update the version number and general metadata in plugins.json as well.

plugins/voxel_shape_generator.js
```diff
@@ -42,9 +42,10 @@ function openMappingsBox(){
                 type: 'select',
                 options: {
                     momap: 'Mojang Mappings',
-                    mcp: 'MCP'
+                    mcp: 'MCP',
+                    yarn: 'Yarn'
                 },
-                default: 'MCP'
+                default: 'Mojang Mappings'
```
JannisX11, line 48: This should point to the key, not the value. So "momap". You can also omit it entirely, that way the first option will be the default.

JannisX11 (changes requested):

Nyfaria:
I have made requested changes

PR #474, opened 2024-02-05
[Cardinal] Fix billboarding, add ability to change scale

ewanhowell5195:
# Billboarding
Fixed the billboarding so that the icons face the screen, rather than face the camera. This caused the icons to noticably rotate, especially when looking down from above:

### Old
image
image

### New
image
image

# Scaling

Added the ability to change the icons scale from settings:
image
image
image


ewanhowell5195:
@Bug1312 since this is your plugin just letting you know about this PR.

PR #487, opened 2024-03-03
Update Figura Format to 0.1.3

KitCat962:


plugins/figura_format/figura_format.js
```diff
@@ -147,266 +115,339 @@
...
+								type: 'checkbox',
+								value: true,
+								nocolon: true
+							};
+							dialog.show();
+						}
+					})
+				}
+			})
+			const bakeIK = new Action('figura_bake_ik', {
```
JannisX11, line 354: This was already in the plugin before this PR, but this seems to duplicate the built-in Bake Inverse Kinematics feature, except no Undo point is being created. Can we remove this?

plugins/figura_format/figura_format.js
```diff
@@ -147,266 +115,339 @@
...
+			})
+			const cycleVertexOrder = new Action('figura_cycle_vertex_order', {
+				name: 'Cycle Vertex Order',
+				icon: 'fa-sync',
+				category: 'edit',
+				condition: { modes: ['edit'], features: ['meshes'], method: () => Format === format && (Mesh.selected[0] && Mesh.selected[0].getSelectedFaces().length) },
+				click() {
+					Undo.initEdit({ elements: Mesh.selected });
+					Mesh.selected.forEach(mesh => {
+						for (const face of mesh.faces) {
```
JannisX11, line 407: Mesh.faces is not an array, it's an object, so the function fails to run.

plugins/figura_format/figura_format.js
```diff
@@ -147,266 +115,339 @@
...
-			let displayFrame = EffectAnimator.prototype.displayFrame
-			EffectAnimator.prototype.displayFrame = function () {
-				if (Format === format) return
-				displayFrame.call(this)
-			}
+						}
+					}).show()
+				}
+			})
+			const cycleVertexOrder = new Action('figura_cycle_vertex_order', {
```
JannisX11, line 399: After testing this, the functionality seems to be different from what was intended.
It does not change the texture mapping or texture rendering as described. For that, you would have to copy all the UV data of the face one vertex over.
It does not reliably invert the face, only in some situations.
The only thing this is always doing is switching the face crease on quads, which is already an option in Blockbench though.

plugins/figura_format/figura_format.js
```diff
@@ -89,14 +56,14 @@
...
 		await_loading: true,
 		creation_date: "2023-07-22",
 		onload() {
-			let particle = EffectAnimator.prototype.channels.particle, sound = EffectAnimator.prototype.channels.sound
-			let format = new ModelFormat('figura', {
+			const particle = EffectAnimator.prototype.channels.particle, sound = EffectAnimator.prototype.channels.sound
+			const format = new ModelFormat('figura', {
 				icon: 'change_history',
 				name: 'Figura Model',
 				description: 'Model for the Figura mod.',
```
JannisX11, line 69: It would be good to change the category here to "minecraft", currently it is labeled as "low_poly".

JannisX11 (commented):
Found a few issues, the rest looks good

plugins/figura_format/figura_format.js
```diff
@@ -147,266 +115,339 @@
...
+			const copyPathAnimation = new Action('figura_copy_path_animation', {
+				name: "Copy Animation Path",
+				description: "Calculates the scripting path to this Animation and copies it to the clipboard.",
+				icon: "fa-clipboard",
+				condition: () => Format === format && (Animation.selected !== null),
+				click() {
+					let path = [Project.name || "modelName", Animation.selected.name]
+					path = path.map(getValidLuaIndex)
+					path.unshift('animations')
+					navigator.clipboard.writeText(path.join(""))
```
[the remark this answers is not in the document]
KitCat962: implimented as suggested

plugins/figura_format/figura_format.js
KitCat962: Yep. Removed

plugins/figura_format/figura_format.js
KitCat962: I just removed it. Somebody a year ago suggested it and nobody else uses it

plugins/figura_format/figura_format.js
KitCat962: welp, this action was removed. Resolved

KitCat962:
I've implemented all the changes you suggested. Anything else I should change?

plugins/figura_format/figura_format.js
```diff
@@ -147,266 +101,280 @@
...
+					Blockbench.showQuickMessage(`Coppied "${pathString}" to the clipboard`)
+				}
+			})
+			const copyPathTextures = new Action('figura_copy_path_texture', {
+				name: "Copy Texture Path",
+				description: "Calculates the scripting path to this Texture and copies it to the clipboard. Results may vary.",
+				icon: "fa-clipboard",
+				condition: () => Format === format && (Texture.selected !== null),
+				click() {
+					let texture = PathModule.parse(Texture.selected.path)
```
JannisX11, line 149: Since the PathModule variable is not defined in the web app, this action should have a condition to only work in the desktop app, or use another approach to process the path, or display an error.

KitCat962: whoops, I completely forgot about the webapp. Fixed

PR #494, opened 2024-03-16
Update Animated Java to 0.4.7

SnaveSutit:
- Added keepNames to ESBuild process to improve debugging errors.
- Added language support for DE.
- Fixed language support for ZH.
- Fix broken icon names.
- Update required BlockBench version to 4.9.4.


JannisX11:
Plugins with min_version 4.8.0 or higher now use the updated repository structure. See https://github.com/JannisX11/blockbench-plugins?tab=readme-ov-file#current-structure.
So when installing this plugin, Blockbench would look for the file in `plugins/animated_java/animated_java.js`.
For this to work, the min_version field needs to be set to the identical value in plugins.json as well.

This also gives you the chance to give the plugin a custom about text and custom icon, although both are optional.
You can also now specify plugin details such as dedicated bug tracker in the plugin meta data. More info: https://github.com/JannisX11/blockbench-plugins/blob/master/schema.json

Regarding the plugin itself, I found some console errors during testing that seem to not be present in the current version of the plugin. Unsure if these errors actually affect any functionality.
image

SnaveSutit:
Oh, sweet. I'll make those changes asap.

Regarding the console errors, it appears as though I merged a funky translation PR without realizing it had a couple small errors in it's implementation. I'll take care of that too

JannisX11:
Cool, just make sure the meta data in plugins.json matches the one in the plugin itself. Currently, the old icon is used in the plugin file, while in plugins.json min_version and await_loading is missing.
By the way, specifying new_repository_format is not required if you are using a min_version of 4.8.0 or higher.

SnaveSutit:
Ooh, that `await_loading` flag is probably why double-clicking to open an ajmodel is causing issues.

PR #506, opened 2024-04-10
Cosmic Reach Model Editor plugin

zHoeshin:
Add the Cosmic Reach Model Editor plugin to the list of Blockbench plugins


plugins.json
```diff
@@ -889,5 +889,17 @@
...
+		"title": "Cosmic Reach Model Editor",
+		"author": "Z. Hoeshin",
+		"icon": "icon.png",
+		"description": "Allows creating, editing, importing and exporting Cosmic Reach block models.",
+		"tags": ["Cosmic Reach"],
+		"version": "1.1.0",
+		"min_version": "4.9.4",
+		"variant": "both",
+		"creation_date": "2024-03-09",
+		"new_repository_format": true,
```
Malik12tree, line 903: The plugin's meta data is mismatched with the meta data defined in the code. (mismatched `tags`, missing `min_version`, `variant`, `creation_date`, `new_repository_format`)
```js
    Plugin.register(id, {
      title: name,
      icon: "icon.png",
      author: "Z. Hoeshin",
      description: "Allows creating, editing, importing and exporting Cosmic Reach block models.",
      tags: [],
      version: "1.1.0",
      creation_date: "2024-03-09",
```

plugins/cosmic_reach_model_editor/cosmic_reach_model_editor.js
```diff
@@ -0,0 +1,287 @@
...
+      icon: "icon.png",
+      author: "Z. Hoeshin",
+      description: "Allows creating, editing, importing and exporting Cosmic Reach block models.",
+      tags: [],
+      version: "1.1.0",
+      creation_date: "2024-03-09",
+      onload() {
+        Codecs.java_block.load_filter.condition = (model) => {
+			return (model.parent || model.elements || model.textures) && (!model.cuboids);
+		}
```
Malik12tree, line 26: You should not directly overwrite the condition for the load filter, as its internal condition is subject to change whether inside Blockbench or an another plugin. You should store the original condition and make use of it inside the overwritten condition.

Define a reference in the plugin's scope.
```js
let originalJavaBlockLoadFilter;
```
```suggestion
        originalJavaBlockLoadFilter = Codecs.java_block.load_filter.condition;
        Codecs.java_block.load_filter.condition = (model) => {
		return originalJavaBlockLoadFilter(model)  && (!model.cuboids);
	}
```
And do not forget onunload to reset the condition..
```js
Codecs.java_block.load_filter.condition = originalJavaBlockLoadFilter;
```

plugins/cosmic_reach_model_editor/cosmic_reach_model_editor.js
```diff
@@ -0,0 +1,287 @@
...
+function sumArrays(a, b){
+    s = []
+    for(let i = 0; i < Math.min(a.length, b.length); i++){
+        s.push(a[i] + b[i])
+    }
+    return s
+}
+
+(() => {
+    let codec, format, export_action, import_action, dialog, properties
```
Malik12tree, line 11: `sumArray` should be defined in the plugin's local scope and not exposed globally.
```suggestion
(() => {
    function sumArrays(a, b){
        s = []
        for(let i = 0; i < Math.min(a.length, b.length); i++){
            s.push(a[i] + b[i])
        }
        return s
    }
    let codec, format, export_action, import_action, dialog, properties
```

Malik12tree (changes requested):

Malik12tree:
Please make use of `originalJavaBlockCond`, To clarify again, the original condition might be changed within Blockbench or by other plugins. Instead, use the saved condition within your new condition. This will protect your plugin from breaking Blockbench or other plugins in the future.

Malik12tree:
Yeah of course! But only to give your plugin more availability in case the original condition (`model.parent || model.elements || model.textures`) changes, instead of hard coding it in your code, just change it to call the original condition. It will behave exactly as it's behaving now, only less subject to break things in the future.

```js
(mode) => {
    return !mode.cuboids && originalJavaBlockCond(model);
}
```

zHoeshin:
@Malik12tree , the issue's long ago fixed

Malik12tree:
@zHoeshin Yeah totally aware. Jannis is the only person that can merge pull requests in this repo, sometimes pull requests may take long if he's busy

JannisX11:
Thanks for your help @Malik12tree !

There are two breaking issure remaining:
* `Codec.java_block` does not exist. Use `Codecs.java_block` instead in the 2 places where it's misspelled
* plugins.json has a trailing comma, as you can seen from the automatic check above

I would also strongly recommend to add icons to your model format and your actions.
Once this is fixed, you may update the creation date field to the current date so it'll show up as new.

JannisX11:
The new icons are not working correctly.
It's a good practice to always test all the changes that you make.
image

PR #511, opened 2024-04-19
Update Animated Java to 0.4.8

SnaveSutit:
- Added temporary support for item components. Thanks @puckisilver!

*The ghost of AJ past haunted the last PR, so we got a priest to perform an exorcism on the file.*


JannisX11:
@SnaveSutit Looks like there is a conflict.

SnaveSutit:
Yeah it's just the .js file itself. It's as simple as overwriting it with the new one

PR #516, opened 2024-04-25
CPM Plugin version: 0.6.16_pre2

tom5454:
Add the CPM Blockbench plugin to the plugins repo.


JannisX11:
I have taken a look at this, the integration is pretty neat.
However, after taking a look at the internal functionings of the plugin, I feel like this repository is not the right place for it. The plugin file exceeds the maximum file size guideline, and also the complexity of bundling a separate modeling tool that is built from a different repository in a different language makes it almost impossible for me to review all code and ensure its safety, especially in regards to future updates.
I've seen that the plugin can already be installed from URL and that's provided on your own wiki. So it should still be fairly easy for users to discover and install it. That would be my preferred way to handle this plugin under these circumstances.

PR #519, opened 2024-04-29
Birth of Bunch o' Screenshots

Fer-Jg:
Hi hi, I just created this plugin, its purpose is to allow users to take Screenshots of multiple models in a single run without having to go through confirmation windows and fixing the posture of every model, making the process as straightforward as possible.

I saw some people (including me) making Java to Bedrock pack conversions and needing sprite creation in a simple way so I think it will be very useful for them.


Fer-Jg:
I messed something up 💀💀💀 will open a different PR

PR #522, opened 2024-04-30
New plugin:

Fer-Jg:
> Hi hi, I just created this plugin, its purpose is to allow users to take Screenshots of multiple models in a single run without having to go through confirmation windows and fixing the posture of every model/camera, making the process as straightforward and quick as possible.
I saw some people (including me) making Java to Bedrock pack conversions and needing multiple sprites creation in a simple way so I think it will be very useful for them.

Fixed previous issues and conflicting change, hope this works now :)


plugins/bunch_o_snapshots/plugin.js
```diff
@@ -0,0 +1,416 @@
...
+        }),
+      };
+    }
+    updateCurrentSettings();
+    Settings.openDialog();
+    Settings.dialog.close(0);
+  }
+
+  // When updating, apply meta data changes to plugins.json entry!
+  BBPlugin.register("plugin", {
```
JannisX11, line 179: Please ensure ID consistency. The plugin directory currently has a different ID compared to the one specified in plugins.json, and the name here plus the file name is just "plugin" when it should be the proper plugin ID.

plugins/bunch_o_snapshots/plugin.js
```diff
@@ -0,0 +1,416 @@
...
+    }
+    BosSettings = null;
+  }
+
+  function updateCurrentSettings() {
+    function invalidDirectoryFallback(fallbackDirectory) {
+      Settings.openDialog();
+      Settings.dialog.close(0);
+      BosSettings["BoSoutputDirectory"].set(fallbackDirectory);
+      Blockbench.showMessageBox({
```
JannisX11, line 48: This error message seems to pop up when first installing the plugin.

plugins/bunch_o_snapshots/plugin.js
```diff
@@ -0,0 +1,416 @@
...
+          }
+
+          let dataUrl = frame.canvas.toDataURL();
+          let img = new Image();
+          img.src = dataUrl;
+          img.className = "allow_default_menu checkerboard";
+          const projectName = Project
+            ? Project.name.replace(/\.geo$/, "")
+            : "screenshot";
+          savePath = `${currentSettings.outputDirectory}\\${projectName}.png`;
```
JannisX11, line 280: The backslash for file paths is only going to work on Windows. Please use the character that works for the respective system. You can use the NodeJS path module (`PathModule`) to make this easier and more reliable.

plugins/bunch_o_snapshots/plugin.js
```diff
@@ -0,0 +1,416 @@
...
+          {
+            name: "Change directory",
+            description:
+              "Open a directory selection window to change your BoSoutputDirectory",
+            category: "BoSetings",
+            value: false,
+            type: "click",
+            click: function () {
+              let dialog = new Dialog({
+                title: "WIP",
```
JannisX11, line 135: This does not seem to do anything, you should comment it out.

plugins/bunch_o_snapshots/plugin.js
```diff
@@ -0,0 +1,416 @@
...
+    Settings.addCategory(
+      "BoSetings",
+      (data = {
+        name: pluginData.title,
+      })
+    );
+    if (BosSettings == null) {
+      doLog("create settings");
+      BosSettings = {
+        BoSdevMode: new Setting("BoSdevMode", {
```
JannisX11, line 103: You can just spell out names with spaces in between, no need to cram it all into one word.

JannisX11 (changes requested):

PR #523, opened 2024-05-02
BBS model exporter: Fixed cube faces not exporting in one texture formats

mchorse:
I saw there was a new folder structure... I'm not sure if I did it right, but I didn't manage to make the same presentation with the icon and about.md work with offline installation.


JannisX11:
min_version should be set to 4.8.0 or later for the new structure to work, see readme.md.
The about.md file should be formatted in markdown. HTML does work to some extend, but markdown is the preferred method.

mchorse:
Ah, thank you! I didn't know about HTML, I saw Ewan's plugin and I assumed that's how it supposed to be.

PR #536, opened 2024-05-13
Fix all plugin metadata inconsistencies

ewanhowell5195:
There are lots of inconsistencies in the plugin metadata compared to the plugins.json file. This pull request fixes them all.


Tslat:
If there's an issue with individual plugins or their meta, I'd prefer you raise an issue with the plugin author so that they can look into it internally

This is just an invisible failure waiting to happen

PR #541, opened 2024-05-22
Azurelib Animator 1.0.7

AzureDoom:
Two changes with this bump: 

- Always enables the item display settings to export with `builtin/entity` as the parent value. It is a hard requirement to display items correctly with Azurelib. The previous check wasn't working 100% of the time, so it's easier to force it. I checked with a vanilla item, and this was correctly only applied to Azurelib Display exports.

- Enables the animated texture support. This works for 1.18+ of Azurelib as long as the texture is a .mcmeta file created by the mod developer. 

If you have any feedback on these changes or if I need to adjust anything, feel free to poke me in Discord as I'm in the BlockBench discord still!


JannisX11:
@AzureDoom Make sure to set has_changelog to true in the plugin meta data in both places, so that the changelog will show up in Blockbench.
New versions in the changelog should be added to the bottom of the file so that they show up in the right order.

AzureDoom:
@JannisX11 changes applied!

PR #542, opened 2024-05-22
Group Exporter

MrCrayfish:
Group Exporter is a plugin that allows designers to export a group from the outliner as a model. Only the cubes inside the group (including sub-groups) will be considered for the exported file. Useful when designing models that are made up of multiple models, or even as a tool to design models in a single project instead of multiple projects (within reasonable context) since you can just export the group, not the whole project.

This plugin is very dependent on the fact `computeCube` has a check right at the start `c.export`. If this was applied across the board, this plugin could theoretically support more formats.
https://github.com/JannisX11/blockbench/blob/42c77fa4737de43686e8925ca0ac54e5d241c5c9/js/io/formats/java_block.js#L29

Example of a use case:
This is a backpack I am designing. It needs to be two separate models for my mod, instead of deleting the elements to only export specific elements, I can just export the group.
image


plugins/group_exporter/about.md
```diff
@@ -0,0 +1,14 @@
+# Group Exporter
```
JannisX11, line 1: No need to include the title here, since it's already displayed in the menu.

plugins/group_exporter/outliner_group_exporter.js
```diff
@@ -0,0 +1,101 @@
...
+                let restrict = actionExportGroup.menu_node.classList.contains("opened");
+                restrictCubes(restrict); 
+                originalExportFunction.call(this);
+                resetCubes(restrict);
+            };
+
+            // Patch the context menu to include export options. Have to do this lazily
+            let patchedMenu = false;
+            originalGroupOpenContext = Group.prototype.showContextMenu;
+            Group.prototype.showContextMenu = function(event) {
```
JannisX11, line 80: There's no need to patch the menu, it wouldn't work if the action is triggered from somewhere else. You can add this to the original Export Group action like this:
```
children() {
    return MenuBar.menus.file.structure.find(menu => menu.id === 'export')?.children;
}
```
For an example of how to implement this, check out https://github.com/JannisX11/blockbench/blob/master/js/display_mode.js#L1697-L1704

MrCrayfish:
I've addressed your feedback, and made changes where necessary. I don't have any further additions, so this PR is ready to go.

JannisX11:
I just found one more issue regarding "Export" being a public function. Do you mind if I fix that?

MrCrayfish:
Yeah, you can make the changes.

PR #545, opened 2024-05-26
feat: PBR Preview Plugin

jasonjgardner:
Adds PBR material support to Blockbench

- Overrides scene materials with a `THREE.MeshStandardMaterial` compiled from the project textures and texture layers.
- Allows adjusting the preview scene exposure and tone mapping
- Adds normal map generation action
- Adds `.texture_set.json` export
- Adds support Bedrock's MER map format: Decodes and exports MERs


JannisX11:
Wow, really cool plugin!
I haven't had enough time yet to thoroughly test everything. But from what I have seen, I'm not expecting any major issues.

There are the issues that I've found:
* You are using the ID "bedrock_entity" in two places, but the correct ID for the format is just "bedrock".
* When using different textures as channels, undoing painting changes on them sometimes does not update the texture in the preview correctly
* Is this a Minecraft-only plugin? To me it seems to have a wider range of applications
  * If it is Minecraft only, let's add a tag "Minecraft", or "Minecraft: Bedrock Edition" while there is no Java support
  * If it isn't, please update the about.md text to reflect this. Bedrock should be Minecraft Bedrock Edition or Minecraft Bedrock for short (at least once, to give context), and for bedrock-exclusive features, such as texture sets and MER maps (I believe those are Minecraft-only?) it should be explained that that's what they are for.
* An entry to `plugins.json` is currently missing

Will continue to review tomorrow...

jasonjgardner:
Thanks for checking it out @JannisX11!

I've updated the `about.md` and `plugins.json`. Also tried to correct some of the display conditions for various actions. So now the labPBR actions should be present in Java projects, and MER/texture set actions should be present in Bedrock projects. I blame my Copilot for weird things like `"bedrock_entity"`

Please take a look at the [`applyPbrMaterial()` function](https://github.com/jasonjgardner/blockbench-plugins/blob/d0babc8237d72449e60521b64774eb34b71b13a3/src/pbr_preview/src/lib/applyPbrMaterial.ts#L11). It's used frequently throughout the plugin. Do you have any suggestions on optimizing it? Creating a new `MeshStandardMaterial` so often is probably incorrect; however, I haven't figured out a better way that works.

Feel free to catch me in Discord too.

JannisX11:
> Do you have any suggestions on optimizing it?
Yes, definitely. It looks like at the moment it is recreating the material for each individual face in the scene. It would already be far more efficient, especially in larger scenes, to first get a list of all used textures, and then only update the material once per texture.
The function is also calling `Canvas.updateAllFaces(texture);` once per face. This function can be slow, and it should only be called once.
And then of course it would be more efficient to create the material once, and if it already exists to just update the respective values and texture maps.
Also, you can ignore the `face.enabled` value. It's specific to a certain format, and not what it sounds like.

Unrelated to this: It would be nice to have a way to see what channel a texture or later is currently assigned to.

jasonjgardner:
> Unrelated to this: It would be nice to have a way to see what channel a texture or later is currently assigned to.

Indeed, it would be 😃

I created a panel component for this feature but I'm having problems with the Vue lifecycle it seems. I've got a thread going in Discord about it: https://discord.com/channels/314078526104141834/1247941848979214406

In the meantime, there is a "Select PBR Channel" menu in the PBR Controls panel which will at least help hunt down the assigned channels.

jasonjgardner:
Sorry @JannisX11! I couldn't help myself. I'll refrain from adding more features for now.

I've fixed the channels panel component issue, as well as some of the logic behind the material brush tool's size, and various other issues. [#2379](https://github.com/JannisX11/blockbench/issues/2379) is a blocker for getting the Undo option to work with the tool.

I've also added support for USDZ exports and created a `.bbmat` codec for importing/exporting PBR materials in Blockbench.

Here are some assets I created for testing. (The `.bbmodel` and `.bbmat` are saved as `.json` just for the sake of attaching them to this comment. The USDZ model is pretty cool in AR if you have an iPhone.)

[example_materials.bbmodel](https://github.com/user-attachments/files/15917556/example_materials.bbmodel.json)
[Tiles.bbmat](https://github.com/user-attachments/files/15917560/Tiles.bbmat.json)

USDZ model: [boombox_usdz.zip](https://github.com/user-attachments/files/15917529/boombox_usdz.zip)

----
> And then of course it would be more efficient to create the material once, and if it already exists to just update the respective values and texture maps.

I haven't been able to pull this off yet. Using `needsUpdate` in THREE.js hasn't worked in my implementations. I _think_ Blockbench may need updates to the `Canvas` class, and changes to some UI template strings, for the most efficient method of updating the materials.

JannisX11:
The PBR channels panel is nice!

There are currently a lot of unwarranted changes in plugins.json that are causing merge conflicts.

I've noticed a small error message when selecting the Neutral Tone Mapping option: `three.min.js:6 THREE.WebGLProgram: Unsupported toneMapping: NaN`

The applyPbrMaterial function still looks extremely inefficient. It is creating a new material for each face of each element in the scene. The Canvas.updateAllFaces function also seems to get called once for every material, so if there are 200 elements in a scene, each element will get its faces updated 200 times.
Creating a new material each time it updates is probably okay, but you can do it once per texture, not once per face.

jasonjgardner:
Aha, thank you! There was an error in my logic for updating the faces. They are no longer redundantly updated.

I've also fixed the merge conflict. The plugins.json file was formatted at some point during a merge. My mistake.

> Creating a new material each time it updates is probably okay

Ideally, the PBR material would also be a `ShaderMaterial`. I think that would play nicer with Blockbench's existing Texture material. It'd also allow for previewing popular shader features, like POM for Java textures. I'm not well-versed enough in shaders to get it to work right now.

JannisX11:
Cool, the performance seems to be okay now for most models.
It looks like you didn't push the commit addressing the plugins.json merge issue. The file in this branch still contains changes to a lot of unrelated plugins in plugins.json.
https://github.com/JannisX11/blockbench/issues/2379 is now fixed so you should be able to address the undo issue.

PR #561, opened 2024-06-14
Texture stitcher: fix downloading and update file structure

mchorse:
I didn't know if you change minimal version of Blockbench to `4.8.0` you had to change the structure of the plugin as well. This should fix plugin loading from File > Plugins > Available.


JannisX11:
Ah, sorry, didn't notice that you had increased the min bb version.
This PR adds a changelog file, but changelogs are not enabled for this plugin atm. Make sure to read README.md, it has a lot of important info.

PR #562, opened 2024-06-15
Texture stitcher: enable change log and sync description in plugins.json

mchorse:
I didn't see that README.md has all of that extra information. Now that I've read it, everything should be fine. Thank you for mentioning it! 👍


JannisX11:
has_changelog needs to be in plugins.json as well.

PR #575, opened 2024-07-11
Scene Recorder 2.0

ewanhowell5195:
Adds new formats, and merges Scene Recorder into the GIF Recorder


JannisX11:
I think you are missing a word in the plugin description.
Otherwise the PR looks good

ewanhowell5195:
fixed

PR #577, opened 2024-07-14
[AWSD Controls] Option to only work when holding RightMouse

caioraphael1:
Overall after being edited:
- New option to only enable the controls if the movement is enabled AND is holding the Right Mouse Button.
  - Off by default
  - Disabling the option returns to the old behavior.
- Creation of 2 new buttons: Move Faster and Move Slower, bound to shift and alt by default, with options to change it in the settings.
- The state of the 'WASD Toggle' tool is now saved across restarts.

image

I really like those new options, as they make the movement more in line with Blender and Godot, for example.

### DIFF
--- a/plugins/wasd_controls.js
+++ b/plugins/wasd_controls.js
@@ -3,6 +3,7 @@
 let deletables = [];
 
 let old_animate;
+let rightMouseDown = false;
 
 BBPlugin.register('wasd_controls', {
 	title: 'WASD Controls',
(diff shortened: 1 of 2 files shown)


plugins/wasd_controls.js
```diff
@@ -3,53 +3,54 @@
...
+let rightMouseDown = false;
 
 BBPlugin.register('wasd_controls', {
 	title: 'WASD Controls',
 	icon: 'sports_esports',
 	author: 'JannisX11',
 	description: 'Adds a WASD controlled viewport navigation mode',
 	about: 'The WASD mode can be enabled from the View menu.\nThe keys can be remapped in the keybindings menu.\nThe sensitivity can be changed in the settings under Preview, along with the movement plane.\nHold Control to move faster.',
-	version: '1.1.0',
+	version: '1.2.0',
```
JannisX11, line 14: Make sure to apply metadata changes like version changes to plugins.json as well.

plugins/wasd_controls.js
```diff
@@ -86,14 +93,21 @@ BBPlugin.register('wasd_controls', {
...
 
 		deletables.push(new Setting('wasd_y_level', {
 			name: 'WASD Navigation at Y Level',
 			description: 'Navigate using WASD at consistent Y level rather than on camera plane',
 			category: 'preview',
 			value: true
-		}))
+		}));
+
+		deletables.push(new Setting('requires_hold_right_mouse', {
```
JannisX11, line 105: Settings are global. The setting name and ID do not make sense outside of the context of this plugin. The context needs to be included, like it is in the other setting.

plugins/wasd_controls.js
```diff
@@ -86,14 +93,21 @@ BBPlugin.register('wasd_controls', {
...
+			name: 'Only works when holding the right mouse button',
+			description: 'The WASD Controls needs to be enabled for this to work.',
+			category: 'preview',
+			value: true
```
JannisX11, line 109: Let's make this false by default to not break the workflow for people who are used to how it works now.

JannisX11 (changes requested):

plugins/wasd_controls.js
caioraphael1: I'm a bit confused, you mean the name of the setting? If that's the case, is 'wasd_requires_hold_right_mouse' a better name?

caioraphael1:
I also just made changes locally regarding:
- Make the zoom work correctly if not pressing the Right Mouse Button while having the 'wasd_requires_hold_right_mouse' enabled.
- Creation of 2 new buttons: Move Faster and Move Slower, bound to shift and alt by default, with options to change it in the settings.
- Setting to modify the speed of the Move Faster and Move Slower modifiers.

I have these changes locally, but I wasn't sure if it was correct to commit them here

JannisX11:
You still need to update the name of the new setting. It will be visible in the main settings menu.
Feel free to include the other changes here as well.

caioraphael1:
> You still need to update the name of the new setting. It will be visible in the main settings menu. Feel free to include the other changes here as well.

Ooh, now I get it. I think now is correct.

caioraphael1:
I also made a new change that captures the mouse while flying, so you don't click something by mistake while moving the mouse. This locks the mouse in the current position and makes it invisible. It's an experimental feature as it was made using a bunch of input overrides to avoid the default behavior. After this commit is reviewed, I think these new experimental changes could be useful, if consider creating an option to make the 'mouse capture' the default behavior for moving the camera in Blockbench as a whole.

caioraphael1 (commented):
Fixed in a previous commit. (I'm kinda new to github)

caioraphael1 (commented):
Reviewed.

PR #590, opened 2024-07-27
Creative Mode Plugin

caioraphael1:
- More information about the plugin can be found in this thread (https://discord.com/channels/314078526104141834/1264973145681559674) on the Blockbench Discord.


JannisX11:
Had a chance to test it now, sorry for the long wait.
The plugin works pretty well for the most part. Here is my feedback:

* I would recommend to add an about.md file to explain how the plugin works. How do users enable the mode, how do they pick blocks, how does navigation work, and how do they exit the mode.
* Placing blocks still has issues in some cases, especially if the cubes are rotated.
* Suggestion: Add middle click to pick blocks like in Minecraft
* The plugin is currently missing an entry in plugins.json. You'll need a min_version of at least 4.8.0 so that it works with the new folder structure. Learn more in the [README](https://github.com/JannisX11/blockbench-plugins?tab=readme-ov-file#blockbench-plugin-repository).

caioraphael1:
I added all the info requested. Sorry for the delay. I don't have any plans to update the plugin right now, as I'm quite busy with some other projects, so I've added a list of known issues and current limitations if someone wants to give it a try. 

I might come back to this plugin in the future, as it gives quite a nice workflow imo, I'm quite proud of it :)

PR #602, opened 2024-08-20
Move "Animation to JsonEA Converter" to "Animation to JSON Converter"

Gaming32:
Switch the "Animation to JsonEA Converter" plugin to "Animation to JSON Converter". The original library isn't really updated anymore, and I have [PRed](https://github.com/neoforged/NeoForge/pull/1476) its functionality into the NeoForge mod loader instead.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -875,14 +875,14 @@
 		"min_version": "4.6.0",
 		"variant": "both"
 	},
-	"animation_to_jsonea": {
-		"title": "Animation to JsonEA Converter",
+	"animation_to_json": {
+		"title": "Animation to JSON Converter",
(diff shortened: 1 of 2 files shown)


Gaming32:
The NeoForge PR has now been merged.

JannisX11:
Looks good. Please keep in mind that by changing the plugin ID, you are creating a new plugin and deleting the old one. Meaning users will have to re-install the plugin. Is that intended?

Gaming32:
I am aware of the need to reinstall, yes, but the old ID is entirely defunct, as the mod referenced in the old ID hasn't been updated in over a year.

PR #614, opened 2024-09-11
Added "Minecraft Image to Blocks Model" plugin.

MarkusBordihn:


JannisX11:
Reviewed the plugin today:
* Undo integration seems to be missing. Importing a PNG, then pressing Undo does not work correctly.
* I noticed that when importing two images of different sizes, the image that was imported first becomes invisible
* Especially if this is intended for the Java block/item format, larger images can generate beyond the 48x48x48 pixel limit, which can result in users having issues when they want to continue editing the model.
* The plugin failed to load in the web app. Consider supporting the web app or setting it to desktop-only.

Correct me if I'm wrong, but generally the only functionality this seems to offer over the built-in Import > Extruded Texture is generating based on pixel color. Which isn't necessarily a blocker, but something to consider.

PR #624, opened 2024-09-28
AzureLib Utils v1.0.10

AzureDoom:
- Auto-convert bedrock format into a format that works properly with the mod.
- Remove item display transforms (added by bedrock) from geo file exported for azure_models
- Tweaked ArmorTemplate Names/pivots
- Rewrote the about.md, removing the gradle info as that's part of the wiki and adding some images to make it prettier. (if you don't want my sponsorship section at the end please let me know and I'll remove it).

If you have any questions, feel free to let me know!


JannisX11:
The changelog has 3 categories with the same title and one change each. That seems unintentional.

AzureDoom:
> The changelog has 3 categories with the same title and one change each. That seems unintentional.

Yup, you are right, fixed now!

PR #628, opened 2024-10-01
Update Double Sided Cubes to v1.1.0

SnaveSutit:
- Removed invisible conditions.
- Cancel inversion if the project has BoxUV enabled, and tell the user why.
- Automatically disable BoxUV on inverted cubes if project UV mode isn't defaulted to BoxUV.
- Use `Cube.selected` instead of `selected` to loop through selected cubes.
- Formatted with prettier.


JannisX11:
There were already meta data inconsistencies with plugins.json in the description, but now the tag list is missing too.

PR #633, opened 2024-10-07
Added Tweaks & Stuff

legopitstop:
**Tweaks & Stuff** adds a handful of utility features that can be toggled in settings. By default, all tweaks are enabled.

This plugin features "tweaks" which are pretty much small plugins that can be enabled and disabled. New tweaks can be added by extending the `Tweaks` class and adding it to the `Tweaks.all` array. You can then override the `onEnable` and `onDisable` methods which behave like onload and onunload in BBPlugin but are called when the user toggles it in settings for convenience.

## Tweaks

- Header Color
- Wrap Tabs
- Image Exporter
- Close Actions

See about.md for more info about each tweak.


legopitstop:
Updated metadata, settings, and actions now display on the plugins page along with a few changes.

PR #635, opened 2024-10-10
Updated BAMO to 0.5.0 to enable 1.20.1 support (and add the animation…

tmudway:
… toggle)

Added a new dropdown box on the first tab to select which version you want to export for. This just changes whether the textures are exported in the block/ or blocks/ folder.

Also includes the first stage of adding animation support, generating the relevant .mcmeta file when toggled. I've kept in mind your suggestion to link to the documentation but I need to get that up to snuff before I start throwing it around beyond the mod team im working with


tmudway:
Updated version ID to avoid duplication

JannisX11:
Please ensure that plugin metadata and metadata in plugins.json matches. Currently plugins.json uses the old icon. Also make sure to point to the plugin icon file ("icon.png") rather than embedding it into the file as base64. This helps optimize the load time when viewing plugins.

tmudway:
> Please ensure that plugin metadata and metadata in plugins.json matches. Currently plugins.json uses the old icon. Also make sure to point to the plugin icon file ("icon.png") rather than embedding it into the file as base64. This helps optimize the load time when viewing plugins.

Fixing now

tmudway:
Changes should be made now. Let me know if there are any lingering issues!

PR #636, opened 2024-10-11
Update format version to 1.21.10 to work with custom components Minecraft Item Wizzard

WalkQuackBack:

Updates the format version to 1.21.10 to work with custom components.


JannisX11:
I can't accept PRs to these plugins since the source is hosted off-site, but I'll consider the changes for the next update.

PR #638, opened 2024-10-22
Update Animated Java to v1.5.2

SnaveSutit:
Changelogs:
https://github.com/Animated-Java/animated-java/releases/tag/v1.5.0
https://github.com/Animated-Java/animated-java/releases/tag/v1.5.1
https://github.com/Animated-Java/animated-java/releases/tag/v1.5.2


JannisX11:
plugins.json entry has some missing fields. At least variant: desktop is required for this plugin, tags are probably also not intentionally removed.
What changed in 1.5? It looks like the changelog that is linked is incomplete.

SnaveSutit:
I implemented a new build script that automatically copies over all changed values, but apparently I had missed version and tags.
Those should be back now.

JannisX11:
Min_version is still missing, without it Blockbench would assume it's a legacy version and try to load the plugin from plugins/animated_java.js.
I also noticed some error messages when installing the plugin for the first time:
image

SnaveSutit:
I've fixed the download issue by switching download APIs, and updated `min_version`. I had `min_blockbench_version` in my package, but was looking for `min_version`, so it was defaulting to undefined.

PR #640, opened 2024-10-26
"Meshy" Fix

Shadowkitten47:
The plugin now uses events rather than recreating the Format.compile and other functions, Also comes with a pretty big bug fix. Fixing the x-axis invetation due to the pretty big changes I waited a ~week for people to find bugs but none.


Shadowkitten47:
Oh I also scraped that bundler thing because now that the code is much smaller it's easier to organize in a single file and works just fine

JannisX11:
Nice that you managed to convert it to event hooks, that should make it easier to maintain.
I experienced several issues with function names not matching when loading the plugin. It looks like some functions are prefixed with "meshy", but they are referenced without that prefix. That's a breaking issue.
I also noticed issues with meshes having incorrect offsets if they had different pivot points. This affected both the in-game preview, and the scene in Blockbench if you re-opened the geometry file.

Original:
image

Reopened:
image

Minecraft: 
image

Shadowkitten47:
Yea, added the prefix last second although I was sure I got them all. I also noticed issues with meshes having incorrect offsets if they had different pivot points. This is an issue in both the current version and past, I just don't know how to fix this issue sense the mesh position is fix to the origin but they also independent I really looked hard into the source but I couldn't find a solution

Shadowkitten47:
@JannisX11 The mesh.pos and mesh.orgin are locked together, but you can very clearly see that you can move them separately as is the issue. It's not the vertices changing but the object. So I need to translate all the vertices according to the objects pos on load a perferable undo it when loaded back into BB. So my question is how do I find how much it's been moved with the move tool vs. origin

Shadowkitten47:
@JannisX11 Fixed it.
It was because I assigned the meshes origin after I did the points which would then shift all the point over. It not does it before assigning the positions. Should work ( aka it worked for me ) now

JannisX11:
Still doesn't work correctly
Here is the test file that I used
[rustycar.geo.zip](https://github.com/user-attachments/files/17787821/rustycar.geo.zip)

Shadowkitten47:
@JannisX11 Did you use entity wizard?

Shadowkitten47:
@JannisX11 Fixed lol

Shadowkitten47:
The test model was very helpful thank you
image

Shadowkitten47:
Please don't be another issue

JannisX11:
Unfortunately I'm still seeing similar issues. Additionally, I'm having the issue that after saving the position of elements in Blockbench moves. It doesn't update visually, but you see that the transform gizmo is in the wrong spot, and once you move it you'll see that it's in the wrong location. Is that not the case on your end?

This is an issue that can happen if you modify vector arrays of the mesh directly instead of creating a copy via `.slice()`.
So if you do something like this, it will modify the mesh in Blockbench and not just the export.
```js
let vertex_position = mesh.vertices[vertex_key];
vertex_position[2] = 5;
```

Shadowkitten47:
No I don't see the position shifting after for me? Strange Let me check that it's correct

Shadowkitten47:
I fixed the issue of it moving after saving. Rotation is hard. but I think was because that BB x and y rotation are flip from Minecraft if I'm wrong tell me but it seemed to fix the issue. Since that's what you do in bedrock.js for texture meshes and locators

PR #654, opened 2024-11-22
FrostCraft Blocker

dysceedsKOTE:
UPLOAD VERSION 1.0.0 
FrostCraft Blocker

The plugin allows you to convert a model consisting of cubic rectangles into a mold made of 1x1x1 blocks with UV mapping like minecraft blocks. The plugin helps create models for resource packs that use textures from Minecraft for models. Simply select a group of meshes or a single mesh and press the split 1x1x1 button in your tools menu.


JannisX11:
Please read https://github.com/JannisX11/blockbench-plugins/blob/master/README.md, there are several things that don't align with the repository guidelines.

PR #670, opened 2025-01-18
BBS exporter: UX update and UV fix

mchorse:
New update for BBS exporter plugin:

* Added load filter to be able to drag and drop `.bbs.json` files to import them much easily
* Added boolean option "Export to folder", which automatically exports both `.bbs.json` file and also any textures in the project to picked folder
* Fixed UV coordinates mismatch when exporting with a texture that has different UV size from texture size


JannisX11:
Didn't test it, I'll leave that up to you. But it looks good as far as I can tell from reading the code.

PR #671, opened 2025-01-18
feat: USD Codec

jasonjgardner:
I've ported the USDZ export feature from my PBR plugin into its own codec plugin. Pretty straightforward.

From Blockbench:
Command block texture in Blockbench
Redstone lamp texture in Blockbench

To Adobe Substance 3D Stager:
Blocks rendered USD scene


JannisX11:
I was having trouble exporting models without PBR material, the elements with default textures were missing.
Also there seems to be no support for bones / hierarchy.
This this is now a standalone plugin, those two caveats should either be resolved, or mentioned in the plugin about page, since people will expect this to work as a general purpose 3D export format.
Additionally, the "await_loading": true option can be removed from the plugin config, since the plugin does not need to be active during startup.
Otherwise looks good!

jasonjgardner:
Re-opening because I still need this feature 😅

PR #678, opened 2025-01-31
Structure importer: rewrite NBT file picker

mchorse:
I'm not really sure if it's up to the standards. I didn't really want to rewrite the way dialog panel, so it's more of a monkey patch pull request. I mainly wanted just to fix the plugin to be able to import structures again in the newest version. 😅 

If you have any tips I could improve it without investing dozens of hours into it, it would be nice!


JannisX11:
Good enough for now, thanks!

PR #695, opened 2025-03-01
[Resource Pack Utilities] Fix glTF, OBJ, FBX, Collada batch export

ewanhowell5195:
Fixed Batch Exporter failing to export glTF, OBJ, FBX, and Collada models


JannisX11:
Plugins.json version bump is missing.

PR #705, opened 2025-03-19
Fix "Group Exporter" not available in Blockbench

MrCrayfish:
I forgot to publish my "Group Exporter" plugin into the `plugins.json` file. Should now be ready to go


JannisX11:
Huh yeah, looks like we missed that in the last PR.
It's a bit redundant now since collections have been added, but could still have its uses for some. I'll merge it.

PR #708, opened 2025-03-24
Fixed GeckoLib plugin not working on web.

ZigyTheBird:
I hope there are no more issues with this build.


Tslat:
🤔 
Interesting that that got missed
Looks alright

PR #710, opened 2025-04-03
Ignoring if format is java_block on the Simplify plugin when simplifying rotation.

ElitoSear:
This would make the models not load at all in Minecraft.

### DIFF
--- a/plugins/simplify.js
+++ b/plugins/simplify.js
@@ -55,7 +55,9 @@ Plugin.register("simplify", {
 							cubes[i].faces.west.uv = Simplify(cubes[i].faces.west.uv, data.roundAmount);
 
 							// rotation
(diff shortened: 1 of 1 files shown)


JannisX11:
Seems like a small and reasonable change. Although checking for `Format.rotation_snap` rather than the format ID would make it apply to more formats that use the same restrictions.
For an update to take effect, the plugin version needs to be increased in both the plugin and in `plugins.json`.
@the-pink-hacker for visibility.

the-pink-hacker:
I ended up doing the fix myself because about 4 years ago I actually made some pretty major changes to the plugin but never released them. I also really needed to clean up my javascript from when I didn't know what I was doing. Thanks though for noticing this issue.

PR #713, opened 2025-04-13
Simplify v1.0

the-pink-hacker:
- Fix issue where not all UVs where simplified.
- Disallow zero for rounding value.
- Disable rounding of rotation when format has `rotation_snap` #710.


plugins/simplify.js
```diff
@@ -1,100 +1,120 @@
...
+        for (let i = 0; i < cubes.length; i++) {
+            // Size
+            if (data.size) {
+                cubes[i].to = simplify_array(cubes[i].to, data.roundAmount);
+                cubes[i].from = simplify_array(cubes[i].from, data.roundAmount);
+            }
+    
+            // UV
+            if (data.uv) {
+                cubes[i].faces.north.uv = simplify_value(cubes[i].faces.north.uv, data.roundAmount);
```
JannisX11, line 53: This will currently set UV to NaN.

plugins/simplify.js
```diff
@@ -1,100 +1,120 @@
...
+    if (data.roundAmount == 0) {
+        Blockbench.showMessage("Failed, Invalid Number.", "center");
+        return;
+    }
+    
+    try {
+        let cubes;
+
+        if (data.selectOnly == true) {
+            cubes = selected;
```
JannisX11, line 35: Will cause errors when elements other than cubes are selected. Use Cube.selected instead.

JannisX11 (changes requested):

PR #714, opened 2025-04-16
Add plugin: Export to Blender

mr0x13f:
Hi all,
This plugin adds an export codec for Blender scene files (.blend).
It works by exporting the project as GLB, then running a Python script through Blender that imports that GLB and saves it as a Blender scene file.
Calling Blender is a bit janky, and finding the Blender executable is even more janky, but it seems to be pretty seamless in the end.
Unlike the other exporters, I added a (fake) progress bar since it takes a lot longer to export.


mr0x13f:
I noticed the routine for finding the Blender path was taking longer than expected. Turns out it's doing about 8000 more filesystem operations than expected. Will push fix soon.

JannisX11:
I forgot to check this during review, but please provide the about text as a separate markdown file, as seen in other plugins. If it only exists in the plugin source file, it will not be displayed before installing the plugin.

PR #721, opened 2025-05-13
MCP Server

jasonjgardner:
This PR adds a [Model Context Protocol server](https://modelcontextprotocol.io/introduction) via [FastMCP](https://github.com/punkpeye/fastmcp) as well as a handful of rudimentary tools to use with it.

Contributions or suggestions on the tools and their `execute` functions are more than welcome.

### Demo

My one-shot attempt at a model which spells out "MCP PR"

You can see there is still some human-in-the-loop activity required/allowed,

https://github.com/user-attachments/assets/ebf25e48-a007-403a-a773-6071bd1e4de9


JannisX11:
Plugins in this repository should not have the ability to execute external code. I have clarified this in the submission guidelines now, but this has already been enforced in the past.
This is due to plugins available through the built-in store requiring a higher level of trust and safety.
I think for this plugin the best approach would be to host it on an external repository.

jasonjgardner:
@JannisX11 I removed the eval function from the plugin. If there is some opposition to including AI abilities in Blockbench, please say so.

PR #725, opened 2025-05-29
Asset Browser

ewanhowell5195:
Asset browser is a browser that lets you view, modify, export, and compare the Minecraft vanilla assets. Both Java edition and Bedrock edition are supported, as well as their respective snapshots and beta versions.

asset_browser


JannisX11:
Nothing blocking, but here is my feedback:
* The Bedrock Edition release version is called Bedrock Release. Name the java equivalent Java Release and Java Snapshot for consistency.
* Bedrock entity models open without loading texture or animations, would be cool if at least textures were supported.
* When opening a java item model, the name of the texture mesh element is the texture data URL.
* Any reason why downloaded packs are saved as jar files? If they are just ZIP encoded they might as well be saved as zip files, which would make more sense to me. Especially for bedrock which doesn't have anything to do with that file type.

Issue #663, opened 2025-01-04
[Meshy]Broken Models 


NYACAT1003:
For some reason my models dont work and i think its because it puts incomplete parts sometimes like wen all the other have 4

							[
								[2470, 30, 39],
								[2471, 2105, 6585]
							],

Shadowkitten47:
Got any other information? I'll look into it but I'm not sure.  If you could send me the .bbmodel that would be quite useful

Also what format version are you using? Because, if it's above 1.19.30 then it also won't work. if that's the case you can fix it by turning off Meshy Meta Data under the settings. It something I've known about but haven't fixed yet

Shadowkitten47:
If not do you have a section with less than 3 points? it would just be an edge I didn't think about those 
It would look something like this, just remove it or if you need it just put a overlapping point for now. I'll add a case for this in a future update
![Image](https://github.com/user-attachments/assets/f7d9adb3-e512-4c5a-be2d-db5502ca6642)

PR #730, opened 2025-06-07
Meshy 1.0.5

Shadowkitten47:
I removed the meta data feature because it caused issue with later format versions "1.19.30" and above. I also fixed the related issues of it not correctly importing without using meta data. Changed the faces to ensure that 4 points are ALWAYS included, this caused issue sometimes when importing with less than 3 points.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1081,7 +1081,7 @@
 		"icon": "diamond",
 		"description": "Enables the use of a meshes in bedrock formats and to export them to Minecraft Bedrock",
 		"variant": "both",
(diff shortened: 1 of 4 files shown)


JannisX11:
Make sure to update the version number in plugins.json.

Shadowkitten47:
Sorry, I always forget to update the number. At least one, [Issue.](https://github.com/JannisX11/blockbench-plugins/issues/663)
The other issues open don't have much information and I don't know if they would resolved as a result

PR #734, opened 2025-06-18
Update optimize.js

KroziTech:
-Fixed elements with rotation angles that are not a multiple of 22.5
-Added support for removing faces that are completely encased by other elemtents even if they have different rotation angles


KroziTech:
> Don't forget to update plugins.json

Done!

PR #736, opened 2025-07-02
Blockbenchy - Adding 3MF support

heyitsdoodler:
This plugin makes it much easier to go from modelling in Blockbench directly to printing


plugins/blockbenchy/icon.png
JannisX11, line 1: The icon is too large, the maximum resolution for plugin icons is 48x48 as seen in readme.md.

JannisX11 (changes requested):
Hey, just reviewed the plugin. I left a few notes. It seems to work well otherwise.
The plugin will need a different name though, it can't just be the name of the program with a Y at the end. I would recommend something more descriptive like export_to_3mf / Export to 3MF.

heyitsdoodler:
Everything should be in order now

PR #750, opened 2025-08-11
Create fov_cameras.js

Arcomit:
I've added a camera plugin that allows the parent group's scaling properties to control the camera's FOV.


JannisX11:
This does not meet some of the submission criteria.
I'm also unsure what the purpose of this is. Camera FOV can already be animated via keyframes. There is no notable change to the plugin other than that, it's just a duplicate.
Aside from the fact that I don't think this change is necessary, changes like this should be made to the existing plugin, rather than creating a copy of it.

PR #758, opened 2025-08-30
Deprecate old animation_utils plugin, create Geckolib plugin and update to 5.0.0-beta1

Tslat:
Bundled because the change is interdependent on each other

Tested on current netlify release of 5.0.0 and downloaded beta instance, working as stable as far as I can tell.

I did it this way because the ultra-generic name of animation_utils has always been useless given it's specifically for Geckolib, and the Geckolib namespace is unclaimed. This should finally swap this over to the proper namespace

Also updated all of the plugin's tooling


Tslat:
Hold, found a couple things

PR #785, opened 2025-10-18
Add Check Invalid Rotation Plugin

PlanesZwalker:
- Detects non-standard rotations in models before export
- Shows warning dialog with element names and suggested fixes
- Auto-fix functionality for rotations to nearest standard values
- User choice to continue export or cancel to fix first
- Supports multiple rotation fix options
- Compatible with Blockbench 4.12.6+ and Minecraft Bedrock standards


plugins/check_invalid_rotation/check_invalid_rotation.js
```diff
@@ -0,0 +1,700 @@
...
+    }
+
+    Plugin.register('check_invalid_rotation', {
+        title: t('title'),
+        author: 'AnnJ',
+        description: t('description'),
+        icon: 'warning',
+        version: '1.0.0',
+        variant: 'both',
+        tags: ['minecraft', 'rotation', 'validation', 'export'],
```
JannisX11: You can specify up to 3 tags. Make sure they are formatted for display in UI, in this case they should start with an upper case letter.

plugins/check_invalid_rotation/check_invalid_rotation.js
```diff
@@ -0,0 +1,700 @@
...
+            
+            if (!hasProcessedChildren) {
+                checkElement(modelData, 'root');
+            }
+        }
+        
+        return nonStandardRotations;
+    }
+    
+    function applyRotationFix(rotation) {
```
JannisX11: This seems to be unused.

plugins/check_invalid_rotation/check_invalid_rotation.js
```diff
@@ -0,0 +1,700 @@
...
+    function refreshScene() {
+            if (Interface && Interface.updateViewport) {
+                Interface.updateViewport();
+            }
+            
+            if (Outliner && Outliner.update) {
+                Outliner.update();
+            }
+            
+            if (Project && Project.update) {
```
JannisX11: Does not exist

plugins/check_invalid_rotation/check_invalid_rotation.js
```diff
@@ -0,0 +1,700 @@
...
+                Project.update();
+            }
+            
+            if (Canvas && Canvas.updateAll) {
+                Canvas.updateAll();
+            }
+            
+            if (Canvas && Canvas.render) {
```
JannisX11: Does not exist

plugins/check_invalid_rotation/check_invalid_rotation.js
```diff
@@ -0,0 +1,700 @@
...
+                Canvas.render();
+            }
+            
+            if (Timeline && Timeline.update) {
```
JannisX11: Does not exist

plugins/check_invalid_rotation/changelog.json
```diff
@@ -0,0 +1,18 @@
+[
```
JannisX11: The changelog file is malformatted.

plugins/check_invalid_rotation/about.md
```diff
@@ -0,0 +1,46 @@
...
+
+## 🔧 Technical Details
+- Compatible with Blockbench 4.12.6+
+- Uses only built-in Blockbench APIs
+- No external dependencies
+- Lightweight and efficient
+- Automatic rotation normalization
+- Smart duplicate detection
+
+*Version 1.0.0 - Compatible with Blockbench 4.12.6+*
```
JannisX11: Version and compatibility is already displayed on the plugin page, I would recomment not having it here again since it's always another step to change it when updating the plugin.

JannisX11 (changes requested):
Left comments on a few issues that I've found.

Apart from that, the issue that I noticed was that no check was performed when saving via Ctrl + S, or File > Save Model.

Btw, in case you didn't see, Minecraft recently lifted the 22.5 degree limit for models, and on bedrock it has never existed in the first place. So I assume this is just intended for legacy versions, any maybe for stylistic consistency if the style requirement is to align with these rotations?

Also, for anything validation-related I can also recomment the Validator API, it can display issues in real time in the UI so could be an addition: https://web.blockbench.net/docs/classes/custom_validator.ValidatorCheck.html

PR #786, opened 2025-10-18
Add Translation Plane Gizmo Plugin

PlanesZwalker:
- Enhanced 2D plane gizmo for simultaneous multi-axis translation
- RGB color coding: XY (Blue), XZ (Green), YZ (Red) planes
- Smart integration with Blockbench's transform system
- Grid snapping support for precise positioning
- Blender-style handles that scale with camera zoom
- Works with cubes, meshes, and groups
- Compatible with Blockbench v5

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1244,6 +1244,16 @@
 		"tags": ["Texture'", "Viewport", "Utility"],
 		"repository": "https://github.com/JannisX11/blockbench-plugins/tree/master/plugins/texture_filtering"
 	},
+	"translation_plane_gizmo": {
+		"title": "Translation Plane Gizmo",
+		"author": "AnnJ",
+		"description": "Enhanced 2D plane gizmo for simultaneous multi-axis translation",
+		"icon": "control_camera",
+		"version": "1.0.0",
+		"min_version": "4.8.0",
+		"variant": "both",
+		"tags": ["Gizmo", "Translation", "Tools"]
+	},
 	"gltf_importer": {
 		"title": "glTF Importer",
 		"author": "0x13F",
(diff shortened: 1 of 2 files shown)


PlanesZwalker:
I've reduced/refactored my code with all the needed improvements. Here is your previous comment on the old code I submitted the first time:

Blockbench.language? Doesn't exist, and never has. Same for Blockbench.off(), Canvas.canvas and lots of others.
Everything is wrapped in a try-catch, which means you can't actually tell if some code is doing what it's meant to do, or failing, or doing nothing at all.
There is a custom localization system even though Blockbench already has that included in most APIs.
Actions are apparently being constructed from HTML nodes manually instead of being created via the API.
That makes it difficult to review the plugin.

In terms of functionality, it generally works, which is nice.
I have noticed a few issues, which need to be fixed before it can be merged:

    The about page is translated to French with the language set to English
    The Disable Plane Gizmo toggle disappears as soon as you switch tabs and cannot be assigned a keybind or added to another toolbar.
    When moving with groups inside and then running Undo, the group offsets are not reverted. For this to work, the groups would need to be tracked in Undo.init.
    When hovering over the gizmo planes, they flicker, as if they were not detecting the mouse hover correctly.
    
    
 I can still modify my code if some parts are not good enough ;)

JannisX11:
I am still having issues with the orientation of the translation planes when moving objects in a rotated space.
To reproduce:
- Create a group and rotate it 60 degrees on one axis
- Inside, create a mesh
- Try and move the mesh in Parent space using the translation planes
- The planes are not oriented correctly and the mouse movement does not correctly translate into the movement of the object

Besides that, this PR is missing an entry in plugins.json. That's required so that the plugin can be installed by users. Although make sure to follow the guidelines in readme.md, such as setting a min_version of at least 4.8.0 to support the new format, and excluding the about text since it's in a separate file.

PR #789, opened 2025-10-22
Added Tesera game plugin

sciner:
https://store.steampowered.com/app/3258010/Tesera/
https://discord.com/invite/8C72nfZ3WR
https://tesera.io
https://www.youtube.com/@tesera_game/videos

### DIFF
--- a/plugins/tesera/tesera.js
+++ b/plugins/tesera/tesera.js
@@ -14,10 +14,10 @@
         min_version: "4.8.0",
     }
 
-    const property_name = 'madcraft'
     const removables = []
     const deletables = []
     const toolbars = []
+    const listeners = []
     const default_value = {flags: [], json: null, material: 'regular'}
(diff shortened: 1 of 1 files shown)


JannisX11 (commented):
Hey there! The game is not released yet, is this intended as an internal tool for the developers behind it?
If so, it would be best to host it internally. There is no need to include it in the public repository if it's just used by a handful of people, it will just mean slower review time.
Plugins can be loaded via the "Load Plugin from URL" option as well if they are hosted on a server and served with the correct media type.
The plugin can still be added to this repository when players can use it to mod the game, if that is planned.

sciner:
Hi!
The game isn’t officially released yet, but it has been available as a playtest version for over a year, and it has already been downloaded by more than 100,000 players.
Many players have started creating plugins already, but working with a single file is quite inconvenient for them. Because of that, having it in the repository would be very helpful, even at this early stage.

PR #794, opened 2025-10-26
Add plugin: Texture downscaler

mr0x13f:
This plugin adds the ability to display textures at a lower resolution non-destructively. 
The lower resolution is also used when exporting the model as glTF.
This is useful when you want to be actively editing a high resolution source image for a texture while displaying it at a lower resolution.


JannisX11:
No, sorry for the wait, just needed to find the time to test it!

PR #825, opened 2025-12-19
Update Geckolib plugin to 4.2.3

Tslat:
Bug fixes yay


JannisX11:
I fixed another bug where animations with sound keyframes didn't play correctly, even in bedrock models with the plugin installed. That's on me because the types for EffectAnimators were incorrect/incomplete.

PR #838, opened 2026-01-16
Add Hytale Avatar Loader plugin

PasteDev:
This plugin allows users to load their complete Hytale avatars directly into Blockbench from their avatar skin JSON files. These JSON files are stored in the Hytale cache folder (%appdata%\Hytale\UserData\CachedPlayerSkins) and contain all the player's avatar customizations including body type, colors, clothing, accessories, and facial features.

The plugin automatically processes all the avatar components including body, hair, eyes, facial features, clothing, and accessories, applying the correct gradient maps and colors based on the player's selections stored in the JSON file.

The main goal here is to give the community an easy and convenient way to get their avatar model ready to use for making thumbnails and other creative projects. Instead of manually setting up each part of the avatar, users just select their avatar JSON file and the plugin handles everything automatically.

The plugin reads the gradient sets from the Hytale assets, applies them to greyscale textures, preserves existing colors like gold accents, and loads everything into Blockbench with the correct UV mapping. It also creates timestamped folders to prevent overwriting when working with multiple avatars.

This should make it much easier for content creators and model makers to work with Hytale avatars.


plugins/hytale_avatar_loader/hytale_avatar_loader.js
```diff
@@ -0,0 +1,1624 @@
...
+            description: 'Loads a complete avatar from a JSON file',
+            icon: AVATAR_BLACK_LOADER_ICON || 'icon-player_head',
+            condition: {formats: ['hytale_character']},
+            click: function() {
+                
+                function expandPath(path) {
+                    if (path.includes('%appdata%')) {
+                        try {
+                            const os = requireNativeModule('os');
+                            const appDataPath = process.env.APPDATA || (os.homedir ? pathJoin(os.homedir(), 'AppData', 'Roaming') : '');
```
JannisX11: Instead of this you can just use "SystemInfo.appdata_directory", it doesn't open a permission dialog.

plugins/hytale_avatar_loader/hytale_avatar_loader.js
```diff
@@ -0,0 +1,1624 @@
...
+                             '**Location:**\n' +
+                             '`%appdata%\\Hytale\\UserData\\CachedPlayerSkins`\n\n' +
+                             'Navigate to this folder and select your avatar JSON file.',
+                    buttons: ['OK']
+                }, () => {
+                    Blockbench.import({
+                        type: 'json',
+                        extensions: ['json'],
+                        multiple: false,
+                        title: 'Select Avatar JSON File'
```
JannisX11: You can set the `startpath` field to the recommended path here so the user doesn't have to open it manually.

plugins/hytale_avatar_loader/icon.png
JannisX11: The icon is currently too large, please follow the guidelines in README.md.

PR #847, opened 2026-02-01
Add PoseKey Plugin

OnirisHub:
I added a PoseKey plugin to Blockbench, which allows adding a key to the entire animated group (customize toolbar > PoseKey).

### DIFF
--- a/RAPPORT_CORRECTIONS.md
+++ b/RAPPORT_CORRECTIONS.md
@@ -1,0 +1,67 @@
+# Rapport des corrections – Plugin PoseKey
+
+## 1. Undo (CTRL+Z) – `posekey.js`
+
+### Note originale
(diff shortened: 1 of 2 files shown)


plugins/posekey.js
```diff
@@ -0,0 +1,103 @@
...
+                    }
+
+                    if (selection_base.length === 0) {
+                        Blockbench.showQuickMessage("⚠️ Select an object!");
+                        return;
+                    }
+
+                    let final_targets = getAllDescendants(selection_base);
+
+                    Undo.initEdit({purpose: 'PoseKey Single'});
```
JannisX11: Undo currently does not work. "purpose" is not a valid property. Instead, you need to init the edit with the empty list of keyframes, and then finish it with all the newly created keyframes.

JannisX11 (changes requested):
Left a few notes.

PR #855, opened 2026-02-22
Minor Hytale Bridge Update

tazercopter:
I updated the hytale plugin (serverside) associated with this to no longer need a custom port, so this bb plugin update just changes the default port to 5520 to match Hytale's however it is still fully compatible with the Hytale plugin as I just listen on both 5520 and 8651 (current default).


tazercopter:
There's also icon and descriptions updates

PR #859, opened 2026-02-28
Add easings plugin

respectZ:
## Description
This plugin allows apply easing to Minecraft: Bedrock animation by manipulating `anim_time_update` rather than per-keyframe basis. Easing can be only applied to animations with default or empty `anim_time_update` and known animation length (not infinite or 0).

There are few flaws for this plugin:
- Needs to be reapplied if either duration or loop mode has been changed.
Because easing require `t` from 0 to 1, and `t = current_time/total_duration`. There's no molang for getting animation length. Same case with loop.
- Back and Elastic easings aren't supported
These easings have out of bound value while Minecraft just clamp it within animation length.
- Blockbench preview seems like not showing bounce easings properly?
I guess this is because it doesn't support playing backwards? I've tested in-game and works perfectly. (Blockbench version was 5.0.7)

Variable names are prefixed and hashed based on animation name, there's shouldn't be conflict or collision between them.

## Preview

https://github.com/user-attachments/assets/33437377-91f9-4bf8-b471-67fdcaf7c21c

There's also `Apply Easings` to manage all current animations easing with modal or dialog.


JannisX11 (changes requested):
Hello, sorry for the long wait, I have been busy.
The plugin and formatting looks generally good, I just found two issues:
* The Apply Easings action fails with a console error when animation controllers are loaded in the project
* The has_changelog property is not set in the plugin file itself, so it doesn't load changelogs when sideloaded

PR #885, opened 2026-04-15
AzureLib Plugin v2.1.0

AzureDoom:
v2.1.0 Update

- Reworked the Animation tab with a full rewrite, moving core functionality to custom logic to resolve animation export issues.
image

- Added support for Bedrock easing directly within the Animation tab.
image

- Adds support for animation includes as per https://moddedmc.wiki/en/project/azurelib/latest/docs/misc/animating/animation_includes
image

- Introduced an early work-in-progress IK Manager, allowing IK-based animation workflows with baking to keyframes (still under active development).


JannisX11:
I had issues getting interpolation to work in this version.

Steps to reproduce:
* Create a new AzureLib Animated Model
* Create a group with a cube inside
* Create an animation
* Add a position keyframe at time 0 with values 0,0,0
* Add another position keyframe later with different values
* Leave Interpolation on the second keyframe on Azure Easing and any of the bottom options

Observed result: When playing the animation, the group does not move, it only moves when positioning the playhead directly over the second keyframe.
It still works fine in other formats.

Off topic, I wanted to let you know about the recently added AnimationCodec, which could be interesting for this plugin, as they provide more flexibility when implementing custom animation formats: https://web.blockbench.net/docs/classes/generated_animations_animation_codec.AnimationCodec.html

AzureDoom:
Ok, i have a fix in the tank for the reported interpolation issue, thankfully was just a visual bug here. I am looking at AnimationCodec here now before I push though.

AzureDoom:
Pushed the required changes now plus added support my includes animation file of the animation files as per: https://moddedmc.wiki/en/project/azurelib/latest/docs/misc/animating/animation_includes @JannisX11

AzureDoom:
Yeah, i'm not sure why moving to your AnimationCodec is killing the validator here. Would love some feedback on this.

JannisX11:
The validator has been updated, and it looks like there is something in the way it's bundled that makes it unable to validate the plugin. So you can ignore it in this case.

AzureDoom:
Perfect than this pr is all done unless you have more feedback

PR #886, opened 2026-04-16
Add No Java Limits plugin

Rajdacz:
It's a simple plugin that allows creating oversized Java models for display entities. It basically just scales down the exported model and generates a command to bring the size back in-game.


JannisX11:
Looks good in terms of functionality.
Please fix the validator issues (which you can see in the "Checks" section above). You will need a min_version of 4.8.0 or greater for the new structure (see README.md). Ensure the metadata is consistent and use tabs for intentation in plugins.json.
I would also recommend to add information about how it works, either in the description, or via about.md file.

Rajdacz:
Thanks for the review. I've applied the requested changes - added `min_version`, changed indentation and added description to plugins.json and added about.md.

PR #887, opened 2026-04-17
Fix: importing existing animation.json file no longer resets frame keys (GeckoLib)

matmat37000:
Fixed an issue where importing an existing animation.json file will cause GeckoLib frame keys to be reset.

Changed the codec to call the GeckoLib own parser instead of the Bedrock one.


JannisX11:
@Tslat

plugins/geckolib/src/ts/codec.ts
```diff
@@ -32,14 +37,23 @@ export const format = new ModelFormat(GECKOLIB_MODEL_ID, {
...
 
 // Override the new project panel to allow customisation
-format.new = function() {
+format.new = function () {
     if (newProject(this))
         return openProjectSettingsDialog();
 }
 
+// Override the animation code parser
+animation_codec.loadFile = function (file, animation_filter) {
```
Tslat: We don't appear to be undoing this monkeypatch when the plugin is unloaded? This just stays there forever, we really should be undoing these so that it's a clean slate when we leave

plugins/geckolib/src/ts/codec.ts
```diff
@@ -32,14 +37,23 @@ export const format = new ModelFormat(GECKOLIB_MODEL_ID, {
...
+    debugger;
+    if (isGeckoLibModel())
```
Tslat: Does this even work at file load time? How does Blockbench know what the model format is before it's even loaded the format?

plugins/geckolib/src/ts/codec.ts
```diff
@@ -390,4 +404,226 @@ export function buildDisplaySettingsJson(options = {}) {
...
 }
 
-export default codec;
\ No newline at end of file
+/**
+ * When the animation file is being loaded into the project
+ * <p>
+ * The project <b><u>may not</u></b> be a GeckoLib project, so check it as necessary
+ */
+function LoadFile(file, exportingAnims) {
```
Tslat: Naming convention does not match styling for the rest of the project (should be loadFile)

Tslat (changes requested):

plugins/geckolib/src/ts/codec.ts
matmat37000: The patch I wrote is crude. After further investigation I found out that I could use your `addMonkeypatch` function to patch `Codecs.bedrock.format.animation_codec`, so it's cleaned up with the rest of the monkeypatchs.
If your custom parser is incompatible with the bedrock one and can't replace it we should be creating a custom AnimationCodec (need update of `blockbench-types` to compile ?).

plugins/geckolib/src/ts/codec.ts
matmat37000: It works because it's not when Blockbench imports the model but rather when it imports the `animation.json` file into the animation panel.
Maybe I didn't fully understand how it works, but from my test, this is when it gets called. This was removed in my last commits in favor of the `addMonkeypatch` and `monkeypatchAnimatorLoadFile`.

plugins/geckolib/src/ts/codec.ts
matmat37000: My bad on this one, I just copied the `monkeypatchAnimatorLoadFile` function from `events.ts` because importing it in `utils.ts` broke everything. I don't know why, I never coded in TypeScript before 😅.

matmat37000:
@Tslat I've made changes to the code. It should be ok, but I'm not a TypeScript dev and it's my first time with the language, so point out any mistakes if needed

Tslat (approved):
Tested as working, no immediately identified issues

PR #893, opened 2026-04-26
Z-Fighting Fixer

TVcraft01:
**Z-Fighting Fixer**
One-click fix for z-fighting on overlapping cube faces, including flat/2D blocks.
Handles any number of overlapping cubes, fixes both sides of every conflict.


JannisX11:
The plugin file itself is missing. Please ensure it is in the PR and that the validator runs successfully (`npm run validate z_fighter_fix`)

PR #897, opened 2026-05-04
✨ Release Brush Tuna

SnaveSutit:
Brush Tuna overhauls the Brush Presets system, providing an improved interface, and several new features.

### Features
- Per-preset pen pressure configuration
- Customizable pen pressure curves
- Keybinds for quickly switching between presets using `[` and `]`
- Lock Alpha option can be toggled per Brush Preset
- Improved preset list in Brush Presets menu

about

## Source
https://github.com/Embody-Games/Brush-Tuna


SnaveSutit:
I am unsure how to fix this issue:
image
As the `name` field is invalid when defining a Plugin:
image

I think the validator might be checking for `name` when it should be looking for `title`?

JannisX11:
Yeah you can remove name, it's not a valid property. There's only title.

PR #900, opened 2026-05-09
Add animation support to gltf_importer

Siydge:
## Additions:
* Add animation importing support to the gltf_importer plugin.
* Fix compounding rotation with groups when importing a gltf model.

Example of compounding rotation on previous versions:
Screenshot_20260510_005412

This specifically occurs and wasn't noticed because Blockbench fails to recognize _group rotation_ until the parts are updated or reloaded, i.e. moving the model around, which forces it to re-read the group rotation, or saving and reloading the project.
By applying rotation to just parts, we both fix the incorrect initial rendering issue, and also fix inaccurate / compounded rotation applied (rotation both applied to group and its descendants)

## Current Bugs:
~Animations don't show/render properly until you move the model, or reload the project. Unsure if there's a method in Blockbench to force this kind of an update, or if its a bug with Blockbench.~
  Fixed this, implemented Canvas.updateAll() to ensure these changes are detected.


Siydge:
Currently fixing an issue with inaccurate rotations once more.

Siydge:
I'm still trying to get things 1:1, it imports mostly right, but some animations aren't correct, or positions / rotations aren't perfect.
I think parts use ZXY and groups use XYZ Eulers? I have no clue.

JannisX11:
@mr0x13f

Siydge:
Animations are now 1:1 for Blockbench -> Blockbench, same with groups rotation, and parts rotation.

plugins/gltf_importer/src/plugin/import_gltf.ts
```diff
@@ -378,29 +466,29 @@ function importTexture(threeMaterial: THREE.MeshStandardMaterial|undefined, opti
...
         let absoluteTexturePath = PathModule.join(PathModule.dirname(options.file.path), cacheKey);
-        bbTexture = new Texture().fromPath(absoluteTexturePath);
+        bbTexture = new (window as any).Texture().fromPath(absoluteTexturePath);
     }
     // TODO: are we sure it cant still be some other stupid thing?
 
     if (bbTexture != undefined) {
         bbTexture.name = threeTexture.name || 'texture',
-        bbTexture.add(false);
+            bbTexture.add(false);
```
mr0x13f: Thought at first this identation was wrong, but I think the issue is that the previous line ends with `,` instead of `;`. Could you fix that? That way the indentation change can be reverted also. I'm surprised this compiled before

plugins/gltf_importer/src/plugin/import_gltf.ts
```diff
@@ -90,77 +119,134 @@ export async function importGltf(options: ImportOptions): Promise<ImportedConten
...
-        group = new Group({
-            name: node.userData.name || node.name || 'group',
-            origin: node.getWorldPosition(new THREE.Vector3()).multiplyScalar(options.scale).toArray(),
-            rotation: eulerDegreesFromQuat(node.getWorldQuaternion(new THREE.Quaternion())).toArray() as ArrayVector3,
+        const groupName = node.userData.name || node.name || 'group';
+        const groupOrigin = currentOrigin.toArray().map(round);
+        // Use 'ZXY' order for groups
+        const groupRotation = eulerDegreesFromQuat(node.quaternion, 'ZYX').toArray().map(round);
+
+        group = new (window as any).Group({
```
mr0x13f: this fails on Blockbench version 4.12.6 beacuse `Group` is only accessible as a global, and not under `window` (somehow?)

I would be ok dropping support for 4.12.6 though. That would require changing the `min_version` in `plugin.ts` and `plugins.json`. I think setting it to `5.0.0` would be fine, it seems to work on that version.

mr0x13f:
I can't do super thorough testing because I don't have a lot of files with group animations to try, but the ones I do have work great! Blender also seems to agree on how the animations should look when importing the same glTF

Siydge:
I'm considering this feature complete, I was able to assist a user with importing a rigged and animated model into Blockbench (By baking the armatures in blender of course) and it imports perfectly.
The Euler's appear to be perfect matches in all of my tests, animations are 1:1, and I am confident in its current state.

Siydge:
Other than that. Which I forgot about...

JannisX11:
Seems to generally work in my test as well. Please fix the validator issue (version mismatch), and clean up the code: You use patterns like `(window as any)`, which should be avoided because it skips type checking. With the exception of maybe "Animation" which tends to conflict with DOM types. There is also a bunch of references to functions that don't exist and never have, such as `Animator.updateContent()` or `Timeline.update()`. Let me know if they are mentioned anywhere so I can investigate, but they shouldn't exist, and Typescript should also flag these as invalid if used correctly. Update blockbench-types to get the most up-to-date types.

PR #906, opened 2026-05-19
Fly Mode — FPS-style viewport navigation

nklbdev:
Fly Mode is an improved FPS-style viewport navigation plugin,
replacing the basic WASD Controls.

### Features
- Hold Right Mouse Button to look around and fly
- Pointer lock while flying - cursor stays captured in the viewport
  for unlimited look rotation
- WASD + Ctrl/Shift movement keys (fully remappable in keybindings)
- Smooth camera rotation with clamped pitch and configurable mouse sensitivity
- Invert vertical axis option
- Adjustable base speed, with Faster/Slower modifier keys
- **Mouse wheel speed control:** scroll while flying to double/halve
  the speed on a power-of-2 scale, with a brief toast showing the
  current multiplier
- All settings available under Preview settings

### Activation
Enable from View menu → Fly Mode toggle, or bind a hotkey.

### Preview

https://github.com/user-attachments/assets/bdd58e3e-08a0-4898-bd0b-ceb1ede04ce3

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -759,6 +759,15 @@
 		"has_changelog": true,
 		"tags": ["Format: Generic Model", "Mesh", "Tool"]
 	},
+	"fly_mode": {
+		"title": "Fly Mode",
+		"icon": "sports_esports",
+		"author": "nklbdev",
+		"description": "Adds a FPS-fly viewport navigation mode",
+		"version": "1.0.0",
+		"min_version": "5.0.0",
+		"variant": "both"
+	},
 	"wasd_controls": {
 		"title": "WASD Controls",
 		"icon": "sports_esports",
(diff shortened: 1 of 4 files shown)


nklbdev:
The plugin validator says it found the about-text in the plugin's JavaScript declaration. But I've already moved the text to the `about.md` file and removed it from the plugin! Perhaps the validator is using files from an older revision of my PR, one that predates the amend commits?

> Error: About text specified in meta data. In format version 4.8 or newer, about text should be in about.md


Hmmm.....


I figured it out:
I can't specify `about` in the plugin's JS file. It should be in the `about.md` file. But in the `plugins.json` file, the `about` should be exactly the same as in the JS file. So, I need to remove it there too.

PR #908, opened 2026-05-19
UV Wrapper — transfer textures and UVs between faces with a drag

nklbdev:
UV Wrapper lets you transfer UVs and textures from a source face onto
target faces by simply clicking and dragging across the model.

### Features
- Alt+Click to pick a source face — orange highlight
- Drag across target faces — UVs and texture are transferred in real time
- Hold Shift to fill an entire flat plane of coplanar faces at once
- Hold Ctrl for planar projection (no fold along edges)
- Real-time green preview of affected faces while dragging
- Modifier keys are fully remappable in the plugin settings
- Undo support — each drag is a single undo step

### Usage
1. Enable the UV Wrapper tool (View → Tools, or Ctrl+Shift+U)
2. Alt+Click a face to set it as the UV source
3. Click and drag across other faces to transfer UVs
4. Use Shift/Ctrl modifiers for advanced projection modes

### Preview

Texture continuation on other faces:

https://github.com/user-attachments/assets/b04e6914-3468-47a8-b123-c4d3dc00ce45

Wrapping texture UV plane around faces planes intersection line to finest texturing! Really helpful when you are working with trimsheets and other tileable textures!

https://github.com/user-attachments/assets/c5a5bc7e-0a45-4498-8800-b407c4566718

Example from real project:

screenshot_2026-05-19_00:07:22

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1520,5 +1520,14 @@
 		"website": "https://svdex.moe",
 		"min_version": "4.8.0",
 		"has_changelog": true
+	},
+	"uv_wrapper": {
+		"title": "UV Wrapper",
+		"icon": "texture",
+		"author": "nklbdev",
+		"description": "Transfer UVs and textures from one face to another in real time",
+		"version": "1.0.0",
+		"min_version": "5.0.0",
+		"variant": "both"
 	}
 }
--- a/plugins/uv_wrapper/about.md
+++ b/plugins/uv_wrapper/about.md
@@ -1,0 +1,26 @@
+# About UV Wrapper
+
+UV Wrapper lets you transfer UVs and textures from one face to another in real time with a simple drag — like wrapping a sticker across the surface of your model.
+
+### How it Works
+
+1. Select the **UV Wrapper** tool from the toolbar (or press **Ctrl+Shift+U**).
+2. Hold **Alt** and click a face to set it as the **source** (orange highlight).
+3. Drag across target faces — they instantly receive the source face's UV layout and texture (green highlight).
+4. The face under the cursor is highlighted in blue.
+
+### Modifier Keys
+
+| Key | Action |
+|---|---|
+| **Alt** (configurable) | Pick the source face |
+| **Shift** (configurable) | Fill a flat plane of coplanar faces at once |
+| **Ctrl** (configurable) | Planar projection — transfer UVs without folding across edges |
+
+All modifier keys can be remapped in **Settings → Edit**.
+
+### Visual Feedback
+
+- 🟠 **Orange** — source face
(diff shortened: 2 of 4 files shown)


nklbdev:
Here is another video from my real project:

https://github.com/user-attachments/assets/afbee31e-191c-4ac5-8845-fe5a9d0c7c76

An important key point about this plugin's greatest value: the UV plane bends not around one of the edges of the faces, but around the line along which the planes of these faces intersect. (The quads' MAIN triangles are used — this is important. The plugin doesn't take quad bending into account. If this is critically necessary, I might rework the plugin, but that's unlikely.)

nklbdev:
And one more question for you, @JannisX11: where can I talk to you about the `*.bbmodel` files import plugin for Godot? I've created a very flexible and versatile plugin for Godot that doesn't require exporting through intermediate formats and supports a ton of options. It even supports skeletons and animation, but it's all still in alpha. I use it daily in game production, along with a few other DIY tools that I could also discuss with you.

JannisX11:
> And one more question for you, @JannisX11: where can I talk to you about the `*.bbmodel` files import plugin for Godot? I've created a very flexible and versatile plugin for Godot that doesn't require exporting through intermediate formats and supports a ton of options. It even supports skeletons and animation, but it's all still in alpha. I use it daily in game production, along with a few other DIY tools that I could also discuss with you.

Ideally with a post in the #bb-feedback forum in the Blockbench discord server.

JannisX11:
Hi, sorry for the late review. I have a few points of feedback:
* The tool may be a bit hard to find for some users, I would recommend adding it to the toolbar (`Toolbars.tools.add(...)`)
* I tried to use it on cubes, and currently it just throws errors there. I think it could be useful on cubes too, but if not it should catch the error and display a message why it doesn't work
* UV faces don't update in the UV editor in real time. Running `updateSelection()` will update it.
* When using the tool and switching to a different project tab, the highlighted face is still visible in the other project
* The modifier settings currently display raw translation keys as descriptions in the settings dialog.

nklbdev:
Thank you so much for your feedback! I'll take it into account and rework the plugin. I use it daily; it's probably the most useful thing I've done since the WASD Controls alternative (Fly Mode).

By the way, I'd also like to mention that I'm preparing a new pull request for Blockbench with a modification to how clicks work. I have suggested something similar before, but it was done completely wrong. The gist is this:

For all tools that can create a selection frame, a click is only registered after releasing the mouse if the selection frame hasn't been created.

Sounds simple, right? But it produces simply amazing results! You can confidently select with a box without worrying about clicking right at the start, selecting another object, and interrupting the selection.

PR #909, opened 2026-05-19
Texture Browser — browse and assign textures from local folders

nklbdev:
## Texture Browser

A panel plugin that lets you browse textures from any local folder and
assign them to selected faces with a single click — no need to manually
import each texture first.

### Features

- **📁 Folder picker** — select any folder on disk; textures are listed
  as thumbnails in a responsive grid
- **🔍 Smart filter** — type a substring (e.g. `brick`) and it
  automatically matches `red_brick.png`, `brick_wall.tga`, etc.
  Full glob patterns (`*.png`, `wall_*`) are also supported.
  By default, if a string does not contain any glob special characters
  (simple pattern), it is treated as enclosed in asterisks (*).
- **✅ One-click assign** — click a thumbnail to instantly apply the
  texture to all selected faces across all selected meshes
- **🔖 Used filter** — native toolbar toggle to show only textures
  currently in use, so you can spot what's already assigned
- **🧹 Orphan cleanup** — optional setting to automatically remove
  textures that become unreferenced after reassignment
- **📐 Configurable thumbnail size** — 10%–200% in Settings, updates
  the grid live
- **📂 Configurable scan depth** — control how many subfolder levels
  to scan (1–10)
- **↩️ Undo-safe** — full undo/redo support including restoration of
  cleaned-up orphan textures

### Technical notes

- Uses Blockbench 5.0 native `Toolbar` + `BarTextInput` + `Action` +
  `Toggle` widgets for all controls
- Minimal Vue — only the texture grid and status messages use a
  template; everything else is Blockbench-native API
- Native APIs accessed via `requireNativeModule()` with proper
  permission prompts
- Panel attaches to the Textures sidebar tab in Edit mode
- Desktop-only (`variant: "desktop"`)

### Preview

Texture list filtering:

https://github.com/user-attachments/assets/e7fd26f4-af55-4f40-ba5f-9f53c5c9f0e9

Texture applying and undo:

https://github.com/user-attachments/assets/d56b68b2-0ee2-49d3-9e60-0e3ec314a112

A similar material panel in the TrenchBroom editor:

https://github.com/user-attachments/assets/2082ea47-7ab7-4772-8097-0d8f0793750b

Example from real project:

screenshot_2026-05-19_00:07:22

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1368,6 +1368,16 @@
 		"has_changelog": true,
 		"repository": "https://github.com/Jatzylap/Java-Block-Sequencer"
 	},
+	"texture_browser": {
+		"title": "Texture Browser",
+		"icon": "image_search",
+		"author": "nklbdev",
+		"description": "Browse textures from a local folder and assign them to selected faces",
+		"version": "1.0.0",
+		"min_version": "5.0.0",
+		"variant": "desktop",
+		"tags": ["Texture", "Viewport", "Utility"]
+	},
 	"texture_downscaler": {
 		"title": "Texture Downscaler",
 		"author": "0x13F",
--- a/plugins/texture_browser/about.md
+++ b/plugins/texture_browser/about.md
@@ -1,0 +1,17 @@
+# About Texture Browser
+
+Texture Browser lets you browse textures from any local folder and assign them to selected faces with a single click — no need to manually import each texture first.
+
+### How it Works
+
+1. Click **Browse** to select a root folder on your disk.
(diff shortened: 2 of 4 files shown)


JannisX11:
Looks fun!
Some feedback:
* Would be nice if it worked for other element types like billboards and cubes as well
* The auto-validator found a small issue with the plugin metadata
* You can also use `Filesystem.pickDirectory()` to pick a directory without asking for native module access

PR #910, opened 2026-05-23
update batch group rename plugin

zzz1999:


zzz1999:
@JannisX11

PR #919, opened 2026-06-12
Add Gradient Studio PRO plugin

OrionTheDragon:


JannisX11:
Hey,
Here is my initial feedback. The plugin has a few issues that need to be fixed:
* The plugin icon is not valid. If you want to use a plugin, you need to use the new repo format, see README.md. The icon is also too large.
* Lots of text in the plugin is in Russian. If you have text that is not translated it must be in English.
* Formatting in plugins.json is inconsistent, please use tabs only.
* In my testing Live Preview seemed to only affect the 2D editor, not the preview in the 3D viewport.

Additionally, we have added new guidelines regarding AI since you created those submission, so please check if those are still followed:
* Plugin submissions may not contain AI generated art, whether as plugin icon, images in the about page, or in other places. Vibe-coded plugins will not be accepted. If coding assistant tools with AI were used, you must retain full understanding of how the code functions.

PR #923, opened 2026-06-17
Changes mesh outlines to be calculated based off of normals

CppJavaCoder:

Switch from scaling the mesh and inverting faces, to expanding the mesh by the vertex normals. This allows for a clean outline around more complex meshes


CppJavaCoder:
@JannisX11 Are you okay if I change how this works on mesh objects?
Here is before my change
CoffeeCupBefore
Here is after
CoffeeCupAfter

PR #930, opened 2026-06-23
🎉 Update Animated Java to v1.10.0

SnaveSutit:
[Release Notes](https://github.com/Animated-Java/animated-java/releases/tag/v1.10.0)


JannisX11:
I found some miscellaneous issues:

In the welcome dialog and some other places, this link is broken: https://animated-java.dev/docs/getting-started/creating-a-blueprint

In the Blueprint Settings, there is an error message that is not translated properly
image

In the permission request dialog, the plugin name is empty. This can happen if the API access is requested before Plugin.register runs.
image

Please check if the plugin metadata matches what's in plugins.json. Usually the validator would check for this. It looks like some fields like repository and bug_tracker are currently missing.

PR #948, opened 2026-07-09
Double Sided Cubes: Add support for Optional Box UV + misc bug fixes

JoshieGemFinder:
**Plugin:** Double Sided Cubes

**Problems:**
1. Double Sided Cubes is not available when Box UVs are the default, even if Per-Face UVs are available and in use.
2. Inverted cubes will use invalid Box UVs if the original cube was using Box UVs.
3. Inverted cubes will sometimes be incorrectly named (e.g. inverting "cube1" will create a cube named "cube2 inverted" instead of "cube1 inverted").
4. Double Sided Cubes has legacy plugin metadata that fails to validate.

**Solutions:**
1. Allowed Double Sided Cubes to run when optional Box UVs are supported by the format
3. Inverted cubes are automatically switched to Per-Face UVs if the original cube was using Box UVs
4. Base the inverted cube's name on the original `cube.name`, instead of the name given by `cube.duplicate()`
5. Updated plugin metadata to be valid.

**Additional Changes:**
* Now places inverted cubes immediately after the original cubes in the outliner, instead of at the bottom of the group.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -92,7 +92,7 @@
 		"variant": "both"
 	},
 	"double_sided_cubes": {
-		"title": "Double Sided Cube Generator",
+		"title": "Double Sided Cubes",
 		"author": "SnaveSutit",
 		"icon": "flip_to_back",
 		"description": "Creates inverted duplicates of the selected cube(s) to allow double-sided rendering in Minecraft: Java Edition.",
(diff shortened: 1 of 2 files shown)


JoshieGemFinder:
@SnaveSutit Sorry for the ping, but I believe you made the original plugin. Do you take issue with any of these changes/bugfixes?  
I'm more than happy to change anything you don't like to fit with your original vision.

SnaveSutit:
Very good, I approve of these changes. Thank you for fixing!

PR #949, opened 2026-07-09
🎉 Release Easing Peasy

SnaveSutit:
Source: https://github.com/SnaveSutit/easing-peasy-blockbench-plugin

# Easing Peasy

A [Blockbench](https://www.blockbench.net/) plugin that adds easing curves to animation keyframes, simplifying the process of creating keyframed animations.

## Features

- 11 easing curves: Linear, Sine, Quad, Cubic, Quart, Quint, Expo, Circ, Elastic, Back, and Bounce.
- Each curve supports In, Out, and In-Out modes.
- Adjustable arguments for curves that support them (elasticity, overshoot, bounciness).
- A dedicated "Keyframe Easing" panel in Animate mode for editing the selected keyframe.
- Reversing keyframes correctly reverses their easing direction too.

## Usage

1. Switch to Animate mode.
2. Select a keyframe with `linear` interpolation.
3. Use the Keyframe Easing panel to choose a curve and mode.


SnaveSutit:
I think you might need to add something that makes the validator work for plugins that use the [`blockbench-patch-manager`](https://www.npmjs.com/package/blockbench-patch-manager) library.

JannisX11:
There you go, it runs now, you can now see your errors.
Svelte was throwing some errors so I just removed all error throwing. And there was an issue with a built-in utility from esbuild, which I've fixed.
You can ignore the error about validate.js obviously

PR #956, opened 2026-07-20
Add addTint plugin

YOHEMAL:
## What changed

- Add the `add_tint` plugin using the current per-plugin repository structure.
- Add an English/Japanese About page with usage instructions and an author link.
- Register matching store metadata in `plugins.json`.
- Update to version 0.4.2 so grayscale conversion only affects UV regions used by selected faces.

## Why

Minecraft Java model JSON supports multiple `tintindex` values, while Blockbench's default editing workflow does not provide a convenient way to assign and preview arbitrary indices. This plugin adds that workflow directly to the UV Editor, a dedicated panel, and the Filter menu.

The original grayscale implementation converted the entire texture referenced by a selected face. Version 0.4.2 maps each selected face's UV bounds to texture pixels so unrelated areas keep their original colors, including on layered and animated textures.

## User impact

Users can assign or remove arbitrary tint indices, edit live preview colors, preview texture/color multiplication, and convert only selected-face UV areas to tint-friendly grayscale. Plugin UI follows Blockbench's Japanese language setting and falls back to English for other languages.

## Validation

- `npm run validate add_tint`
- `node --check plugins/add_tint/add_tint.js`
- `plugins.json` JSON parse check
- `git diff --check`

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1544,6 +1544,16 @@
 		"variant": "both",
 		"tags": ["Texture", "Paint", "Tool"]
 	},
+	"add_tint": {
+		"title": "addTint",
+		"author": "YOHEMAL",
+		"description": "Adds multiple Minecraft tint indices and tint color previews.",
+		"icon": "format_color_fill",
+		"version": "0.4.1",
+		"min_version": "5.0.0",
+		"variant": "both",
+		"tags": ["Minecraft: Java Edition", "Utility"]
+	},
 	"easings": {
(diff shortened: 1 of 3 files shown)


YOHEMAL:
Hi! Just following up on this PR since it has been open for about a month.
The plugin is still working with the current Blockbench version and all validation checks pass.
Please let me know if there are any changes needed before review. Thanks!

PR #963, opened 2026-08-01
Fix glTF interleaved attribute import

### LINKED ISSUE #962: glTF importer corrupts interleaved vertex buffers
### Summary

The glTF importer corrupts meshes whose vertex attributes use an interleaved buffer view. Imported geometry appears as exploded or stretched triangles, and UVs are also scrambled.

@mr0x13f

### Steps to reproduce

1. Import a valid glTF 2.0 asset where `POSITION`, `NORMAL`, and `TEXCOORD_0` share one buffer view with `byteStride: 32`.
2. The bundled `GLTFLoader` creates `THREE.InterleavedBufferAttribute` instances for these attributes.
3. Complete the import into a Blockbench generic model.

### Expected behavior

The editable Blockbench mesh matches the glTF geometry and UVs.

### Actual behavior

Most vertices are read from normal/UV slots instead of position slots, producing long spikes and scrambled faces.

### Root cause

`importMeshPrimitives()` reads attribute storage as tightly packed arrays:

```js
position.array[vertexIndex * 3]
uv.array[vertexIndex * 2]
```

For an `InterleavedBufferAttribute`, `.array` is the shared interleaved storage and these offsets ignore both the attribute offset and buffer stride. The loader already exposes stride-safe `getX()`, `getY()`, and `getZ()` methods.

Index access should also use `geometry.getIndex().getX()` for consistency across BufferAttribute implementations.

### Proposed fix

Read position, UV, and index components through the BufferAttribute getter API rather than indexing `.array` directly. This supports both tightly packed and interleaved glTF buffers.

I have a focused fix and regression coverage ready for a pull request.

krapnikkk:
Fixes #962

## What changed
- Read vertex positions through `BufferAttribute.getX/getY/getZ`.
- Read UVs and indices through attribute getters.
- Bump `gltf_importer` to 1.2.1 and rebuild the committed bundle.

## Root cause
`GLTFLoader` creates `InterleavedBufferAttribute` instances for glTF accessors that share a strided buffer view. Directly indexing `.array` assumes tightly packed attributes and ignores the interleaved stride and accessor offset, which corrupts vertex positions and UVs. The attribute getters handle both regular and interleaved buffers.

## Validation
- `npm run build` in `plugins/gltf_importer/src`
- `npm run validate gltf_importer`
- `git diff --check`

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1366,7 +1366,7 @@
 		"description": "Import .GLTF and .GLB models",
 		"icon": "icon.png",
 		"creation_date": "2025-09-25",
-		"version": "1.2.0",
+		"version": "1.2.1",
 		"variant": "desktop",
 		"min_version": "4.12.6",
 		"has_changelog": false,
--- a/plugins/gltf_importer/gltf_importer.js
+++ b/plugins/gltf_importer/gltf_importer.js
@@ -2763,10 +2763,11 @@
     let primitiveToUniqueVertexIndices = [];
     for (let [primitive, primitiveIndex] of valuesAndIndices(primitives)) {
       primitiveToUniqueVertexIndices[primitiveIndex] = [];
-      for (let vertexIndex = 0; vertexIndex < primitive.geometry.attributes.position.count; vertexIndex++) {
-        let x = primitive.geometry.attributes.position.array[vertexIndex * 3];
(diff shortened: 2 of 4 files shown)


JannisX11:
@mr0x13f Please verify and approve these changes to your plugin

mr0x13f:
looks good

